PR #2, opened 2021-09-09
Add support for configuring the plugin

mprahl:
The configuration comes from an input PolicyGenerator configuration
YAML file.


internal/plugin.go
```diff
@@ -0,0 +1,227 @@
...
+		return
+	}
+
+	// Set defaults to the defaults that aren't overridden
+	if p.PlacementBindingDefaults.Name == "" && len(p.Policies) == 1 {
+		p.PlacementBindingDefaults.Name = "binding-" + p.Policies[0].Name
+	}
+
+	if p.PolicyDefaults.Categories == nil {
+		p.PolicyDefaults.Categories = []string{"CM Configuration Management"}
```
ChunxiAlexLuo, line 95: Just think if we need to hardcode each default value here. Maybe define a default struct and use it here?

All rest LGTM

ChunxiAlexLuo: /LGTM

dhaiducek (approved):

PR #3, opened 2021-09-09
Use a package-level default variable to contain all policy defaults

mprahl:
This is to address a comment left here:
https://github.com/open-cluster-management/policy-generator-plugin/pull/2#pullrequestreview-750674862


ChunxiAlexLuo (approved):
/LGTM

PR #5, opened 2021-09-10
Add support for creating placement rules and placement bindings

mprahl:


internal/plugin.go
```diff
@@ -297,3 +345,175 @@ func (p *Plugin) createPolicy(policyConf *policyConfig) error {
...
+	p.outputBuffer.Write([]byte("---\n"))
+	p.outputBuffer.Write(ruleYAML)
+
+	return
+}
+
+// createPlacementBinding creates a placement binding for the input placement rule and policies by
+// writing it to the policy generator's output buffer. An error is returned if the placement binding
+// cannot be created.
+func (p *Plugin) createPlacementBinding(
```
dhaiducek, line 478: What do you think of updating this behavior so that we have a single PlacementBinding for the entire set of policies? That can always be a later enhancement, though...

internal/plugin.go
```diff
@@ -297,3 +345,175 @@ func (p *Plugin) createPolicy(policyConf *policyConfig) error {
...
+// If the placement rule name is in the skip map and is set to true, it will not be added to the
+// policy generator's output buffer. An error is returned if the placement rule cannot be created.
+func (p *Plugin) createPlacementRule(policyConf *policyConfig, skip map[string]bool) (
+	name string, err error,
+) {
+	plrPath := policyConf.Placement.PlacementRulePath
+	var rule map[string]interface{}
+	// If a path to a placement rule is provided, find the placement rule and reuse it.
+	if plrPath != "" {
+		name, rule, err = p.getPlrFromPath(plrPath)
```
dhaiducek, line 414: Nice--I see the update to output the PlacementRule. Thanks! 😄

dhaiducek (commented):
I'll give the flows another look after lunch, but this looks good to me so far.

internal/plugin.go
mprahl: @dhaiducek so there is a placement binding created per placement rule since a placement binding can only accept a single placement rule. The only way to get a single placement binding with the code in this PR is to set `placementRulePath` on each policy to the same placement rule.

I think we could enhance the code so that if there are multiple policies that specify the same cluster selector, we could create just a single placement rule and thus a single placement binding.

dhaiducek: Yes, or as long as a policy isn't overriding the placement in `policyDefaults`, we could probably go ahead and consolidate them too.

dhaiducek (approved):

internal/plugin.go
dhaiducek: I'm also wondering about the requirement to have `placementBindingDefaults.name`. It seems we could alternatively allow a placement suffix (or prefix) and tack it to the policy name to create the name of the binding, similar to how we handle the rule. (That's what the UI does currently.)

PR #7, opened 2021-09-10
Consolidate placement rules based on cluster selectors

mprahl:
In order to know what to call the common placement rule, the new fields of
policyDefaults.placement.name and policies[0].placement.name are added.
If neither of these are set, then no placement rule consolidation will
occur.

Additionally, this now checks for duplicate placement rule names and
returns an error when that occurs.


dhaiducek (approved):
LGTM! Makes the logic a bit more complex, but I like the results.

PR #9, opened 2021-09-14
Make default binding name optional

dhaiducek:
Rather than running an initial validation on the `placementBindingDefaults.name`, this moves the logic deeper, leveraging the new placement flows to determine whether a default binding name is strictly necessary.

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -147,8 +147,8 @@ func (p *Plugin) Generate() ([]byte, error) {
 			policyConfs = append(policyConfs, &p.Policies[i])
 		}
 
-		// If there is more than one policy but no default binding name specified,
-		// throw an error
+		// If there is more than one policy associated with a placement rule but no default binding name
+		// specified, throw an error
 		if len(policyConfs) > 1 && p.PlacementBindingDefaults.Name == "" {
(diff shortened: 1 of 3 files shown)


mprahl:
@dhaiducek this looks good. Could you please rebase on the latest `main` branch and add a unit test for this?

dhaiducek:
Yes! I'll work on a unit test to cover the other PR also.

internal/plugin.go
```diff
@@ -141,20 +141,35 @@ func (p *Plugin) Generate() ([]byte, error) {
...
-		plcBindingCount++
 		// Determine which policies to be included in the placement binding.
 		policyConfs := []*policyConfig{}
 		for _, i := range policyIdxs {
 			policyConfs = append(policyConfs, &p.Policies[i])
 		}
 
-		// If there are multiple policies, still use the default placement binding name
-		// but append a number to it so it's a unique name.
+		// If there is more than one policy but no default binding name specified,
```
mprahl, line 150:
```suggestion
		// If there is more than one policy associated with a placement rule, but no default binding name specified,
```

mprahl (approved):

PR #13, opened 2021-09-16
Add-option-to-not-consolidate-under-single-ConfigurationPolicy-#15836

ChunxiAlexLuo:
Signed-off-by: Chunxi Luo chuluo@redhat.com

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -21,6 +21,10 @@ policyDefaults:
   # annotation. This defaults to ["CM-2 Baseline Configuration"].
   controls:
     - "CM-2 Baseline Configuration"
+  # Optional. Default value true will consolidate multi manifests 
+  # specified in policyConf into single policy template. 
+  # Set to false will skip this consolidation.
+  consolidatedManifests: true
   # Optional. When the policy references a Kyverno policy manifest, this determines if an additonal
   # configuration policy should be generated in order to receive policy violations in Open Cluster
   # Management when the Kyverno policy has been violated. This defaults to true.
@@ -91,6 +95,8 @@ policies:
     disabled: false
     # Optional. (See policyDefaults.informKyvernoPolicies for description.)
     informKyvernoPolicies: true
+    # Optional. (See policyDefaults.consolidatedManifests for description.)
+    consolidatedManifests: true
     # Optional. Determines the list of namespaces to check on the cluster for the given manifest. If
     # a namespace is specified in the manifest, the selector is not necessary. This defaults to no
     # selectors.
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -21,10 +21,11 @@ policyDefaults:
   # annotation. This defaults to ["CM-2 Baseline Configuration"].
   controls:
     - "CM-2 Baseline Configuration"
-  # Optional. Default value true will consolidate multi manifests 
-  # specified in policyConf into single policy template. 
-  # Set to false will skip this consolidation.
-  consolidatedManifests: true
+  # Optional. This determines if a single configuration policy should be
+  # generated for all the manifests being wrapped in the policy.
+  # If set to false, a configuration policy per manifest will be generated.
+  # This defaults to true.
+  consolidateManifests: true
   # Optional. When the policy references a Kyverno policy manifest, this determines if an additonal
   # configuration policy should be generated in order to receive policy violations in Open Cluster
   # Management when the Kyverno policy has been violated. This defaults to true.
@@ -95,8 +96,8 @@ policies:
     disabled: false
     # Optional. (See policyDefaults.informKyvernoPolicies for description.)
     informKyvernoPolicies: true
-    # Optional. (See policyDefaults.consolidatedManifests for description.)
-    consolidatedManifests: true
+    # Optional. (See policyDefaults.consolidateManifests for description.)
+    consolidateManifests: true
     # Optional. Determines the list of namespaces to check on the cluster for the given manifest. If
     # a namespace is specified in the manifest, the selector is not necessary. This defaults to no
     # selectors.
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -74,7 +74,7 @@ func (p *Plugin) Config(config []byte) error {
 	return p.assertValidConfig()
 }
 
-// Generate generates the policies, placement rules, and placement bindings and returns them as
+// Generate the policies, placement rules, and placement bindings and returns them as
 // a single YAML file as a byte array. An error is returned if they cannot be created.
 func (p *Plugin) Generate() ([]byte, error) {
 	// Set the default empty values to the fields that track state
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -74,7 +74,7 @@ func (p *Plugin) Config(config []byte) error {
 	return p.assertValidConfig()
 }
 
-// Generate the policies, placement rules, and placement bindings and returns them as
+// Generate generates the policies, placement rules, and placement bindings and returns them as
 // a single YAML file as a byte array. An error is returned if they cannot be created.
 func (p *Plugin) Generate() ([]byte, error) {
 	// Set the default empty values to the fields that track state
@@ -209,20 +209,19 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 		p.PolicyDefaults.Controls = defaults.Controls
 	}
 
-	// Policy expanders default to true unless explicitly set in the config.
-	// Gatekeeper policy expander policyDefault
-	igvValue, set := getDefaultBool(unmarshaledConfig, "informGatekeeperPolicies")
-	if set {
-		p.PolicyDefaults.InformGatekeeperPolicies = igvValue
-	} else {
-		p.PolicyDefaults.InformGatekeeperPolicies = true
-	}
-	// Kyverno policy expander policyDefault
-	ikvValue, set := getDefaultBool(unmarshaledConfig, "informKyvernoPolicies")
-	if set {
+	// Defaults to true unless explicitly set in the config.
+	ikvValue, setIkv := getDefaultBool(unmarshaledConfig, "informKyvernoPolicies")
+	if setIkv {
 		p.PolicyDefaults.InformKyvernoPolicies = ikvValue
 	} else {
 		p.PolicyDefaults.InformKyvernoPolicies = true
+	}
+
+	consolidatedValue, setConsolidated := getDefaultBool(unmarshaledConfig, "consolidatedManifests")
+	if setConsolidated {
+		p.PolicyDefaults.ConsolidatedManifests = consolidatedValue
+	} else {
+		p.PolicyDefaults.ConsolidatedManifests = true
 	}
 
 	if p.PolicyDefaults.RemediationAction == "" {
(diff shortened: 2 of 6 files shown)


internal/types/types.go
```diff
@@ -19,21 +19,24 @@ type PlacementConfig struct {
...
 	// This is named Placement so that eventually PlacementRules and Placements will be supported
-	Placement         PlacementConfig `json:"placement,omitempty" yaml:"placement,omitempty"`
-	RemediationAction string          `json:"remediationAction,omitempty" yaml:"remediationAction,omitempty"`
-	Severity          string          `json:"severity,omitempty" yaml:"severity,omitempty"`
-	Standards         []string        `json:"standards,omitempty" yaml:"standards,omitempty"`
+	Placement             PlacementConfig `json:"placement,omitempty" yaml:"placement,omitempty"`
+	RemediationAction     string          `json:"remediationAction,omitempty" yaml:"remediationAction,omitempty"`
+	Severity              string          `json:"severity,omitempty" yaml:"severity,omitempty"`
+	Standards             []string        `json:"standards,omitempty" yaml:"standards,omitempty"`
+	InformKyvernoPolicies bool            `json:"informKyvernoPolicies,omitempty" yaml:"informKyvernoPolicies,omitempty"`
```
ChunxiAlexLuo, line 35: Here moved two bool positions due to lint complaint: `struct of size 264 bytes could be of size 256 bytes`

internal/utils_test.go
```diff
@@ -106,6 +109,67 @@ data:
 		}
 		assertEqual(t, kind2, "ConfigMap")
 	}
+
+	// test NonConsolidated = true case
```
ChunxiAlexLuo, line 113: TestGetPolicyTemplate is too long and has too many duplicated codes, let me know if we need to subtract a helper and short it here.

internal/plugin.go
```diff
@@ -74,7 +74,7 @@ func (p *Plugin) Config(config []byte) error {
 	return p.assertValidConfig()
 }
 
-// Generate generates the policies, placement rules, and placement bindings and returns them as
+// Generate the policies, placement rules, and placement bindings and returns them as
```
mprahl, line 77: Actually, the comment was intended as it was because it's customary to start a function/method comment with the name of the function/method. It just sounds a bit awkward in this case.

internal/types/types.go
mprahl: Sounds good. Could you please order the bool fields in alphabetical order?

internal/types/types.go
```diff
@@ -19,21 +19,24 @@ type PlacementConfig struct {
...
+	Disabled              bool            `json:"disabled,omitempty" yaml:"disabled,omitempty"`
+	// NonConsolidated = false (default value) will consolidate multi manifests specified
+	// in policyConf into single policy template
```
mprahl, line 38: Instead of documenting this here, could you please document this here like the rest of them:
https://github.com/open-cluster-management/policy-generator-plugin/blob/main/docs/policygenerator-reference.yaml

It's nice to have the comment in the code here but I'd rather not duplicate the explanation, so it'd be preferred to move it to the above linked reference file.

internal/utils.go
```diff
@@ -86,9 +86,11 @@ func getManifests(policyConf *types.PolicyConfig) ([]map[string]interface{}, err
 	return manifests, nil
 }
 
-// getPolicyTemplates generates the policy templates for the ConfigurationPolicy manifests
```
mprahl, line 89: Please keep this comment structure that starts with a short sentence describing the function that starts with the function name. Your other comments can go after it.

internal/utils.go
```diff
@@ -86,9 +86,11 @@ func getManifests(policyConf *types.PolicyConfig) ([]map[string]interface{}, err
...
-// that includes the manifests specified in policyConf. An error is returned
-// if one or more manifests cannot be read or are invalid.
+// policyConf.NonConsolidated = false (default value) will generate a policy templates array
```
mprahl, line 89:
```suggestion
// policyConf.NonConsolidated = false (default value) will generate a policy templates slice
```

internal/utils.go
```diff
@@ -86,9 +86,11 @@ func getManifests(policyConf *types.PolicyConfig) ([]map[string]interface{}, err
...
+// that just one item includes all consolidated manifests specified in policyConf.
+// policyConf.NonConsolidated = true will generate a policy templates array
```
mprahl, line 91:
```suggestion
// policyConf.NonConsolidated = true will generate a policy templates slice
```

internal/utils_test.go
mprahl: It's a nice idea, but I don't want to block this PR on that change.

Feel free to work on this in a separate PR.

internal/utils.go
mprahl: Here is a reference by the way:
https://go.dev/blog/godoc

internal/utils.go
```diff
@@ -101,14 +103,51 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
+}
+
+// void function to set namespaceSelector by reference.
+func setNamespaceSelector(policyConf *types.PolicyConfig, policyTemplate *map[string]map[string]interface{}) {
+	if policyConf.NamespaceSelector.Exclude != nil || policyConf.NamespaceSelector.Include != nil {
+		(*policyTemplate)["objectDefinition"]["spec"].(map[string]interface{})["namespaceSelector"] = policyConf.NamespaceSelector
+	}
+}
+
+func buildPolicyTemplatesItem(policyConf *types.PolicyConfig, objectTemplates *[]map[string]interface{}) *map[string]map[string]interface{} {
```
mprahl, line 150: Could you please add a comment describing the function in this style?
https://go.dev/blog/godoc

ChunxiAlexLuo: Sure, will do all above

internal/utils.go
```diff
@@ -101,14 +103,51 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
+	}
+
+	// check the enabled expanders and add additional policy templates
+	expandedPolicyTemplates := handleExpanders(manifests, policyConf)
+	policyTemplates = append(policyTemplates, expandedPolicyTemplates...)
+
+	return policyTemplates, nil
+}
+
+// void function to set namespaceSelector by reference.
```
mprahl, line 143:
```suggestion
// setNamespaceSelector sets the namespace selector, if set, on the input policy template
```

internal/types/types.go
```diff
@@ -19,21 +19,24 @@ type PlacementConfig struct {
...
+	NonConsolidated bool `json:"nonConsolidated,omitempty" yaml:"nonConsolidated,omitempty"`
```
mprahl, line 39: Could you please also add this to `PolicyDefaults` and handle this in `applyDefaults`? That way it's consistent with other options in the policy generator.

ChunxiAlexLuo:
// cc @mprahl  

https://github.com/open-cluster-management/policy-generator-plugin/pull/13/commits/d175f9ba48e55dfef1d125ffc87f4e31b25de9fd

Will do squash merge later

docs/policygenerator-reference.yaml
```diff
@@ -21,6 +21,10 @@ policyDefaults:
   # annotation. This defaults to ["CM-2 Baseline Configuration"].
   controls:
     - "CM-2 Baseline Configuration"
+  # Optional. Default value true will consolidate multi manifests 
+  # specified in policyConf into single policy template. 
+  # Set to false will skip this consolidation.
```
mprahl, line 26: Since this is user facing documentation, I think we should drop the usage of `policyConf` here since it's a Go type in our project and not known to the user. How about something like this?

```suggestion
  # Optional. This determines if a single configuration policy should be
  # generated for all the manifests being wrapped in the policy.
  # If set to false, a configuration policy per manifest will be generated.
  # This defaults to true.
```

docs/policygenerator-reference.yaml
```diff
@@ -21,6 +21,10 @@ policyDefaults:
...
+  consolidatedManifests: true
```
mprahl, line 27: Could you make the option name an action/verb to match the rest of the options? I'm sorry I didn't catch this previously.
```suggestion
  consolidateManifests: true
```

internal/utils.go
```diff
@@ -87,8 +87,11 @@ func getManifests(policyConf *types.PolicyConfig) ([]map[string]interface{}, err
 }
 
 // getPolicyTemplates generates the policy templates for the ConfigurationPolicy manifests
-// that includes the manifests specified in policyConf. An error is returned
-// if one or more manifests cannot be read or are invalid.
+// policyConf.ConsolidatedManifests = true (default value) will generate a policy templates slice
+// that just one template includes all consolidated manifests specified in policyConf.
```
mprahl, line 91:
```suggestion
// that just has one template which includes all the manifests specified in policyConf.
```

internal/utils.go
```diff
@@ -101,14 +104,52 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 		)
 	}
 
-	objectTemplates := make([]map[string]interface{}, 0, len(manifests))
+	policyTemplatesLength := 1
+	if !policyConf.ConsolidatedManifests {
+		policyTemplatesLength = len(manifests)
+	}
+	objectTemplates := make([]map[string]interface{}, 0, policyTemplatesLength)
```
mprahl, line 111: I think we still want the size to be `len(manifests)` if the `policyConf.ConsolidatedManifests` is `true`.

ChunxiAlexLuo: Correct, thanks for catching this

internal/utils.go
```diff
@@ -102,13 +105,51 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
+	//  just build one policyTemplate by using the above consolidated objectTemplates
+	if policyConf.ConsolidateManifests {
+		policyTemplate := buildPolicyTemplate(policyConf, &objectTemplates)
+		setNamespaceSelector(policyConf, policyTemplate)
+		policyTemplates = append(policyTemplates, *policyTemplate)
+	}
+
+	// check the enabled expanders and add additional policy templates
+	expandedPolicyTemplates := handleExpanders(manifests, policyConf)
+	policyTemplates = append(policyTemplates, expandedPolicyTemplates...)
```
ChunxiAlexLuo, line 139: https://go.dev/blog/slices-intro.  // allocate double what's needed, for future growth.

Just checked that it will double the allocation like (len(policyTemplates) + len(expandedPolicyTemplates)) * 2 here

internal/utils.go
```diff
@@ -102,13 +105,51 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	}
 
 	objectTemplates := make([]map[string]interface{}, 0, len(manifests))
+	policyTemplatesLength := 1
+	if !policyConf.ConsolidateManifests {
+		policyTemplatesLength = len(manifests)
+	}
+	policyTemplates := make([]map[string]map[string]interface{}, 0, policyTemplatesLength)
```
mprahl, line 112:
```suggestion
	objectTemplatesLength := len(manifests)
	policyTemplatesLength := 1
	if !policyConf.ConsolidateManifests {
		policyTemplatesLength = len(manifests)
		objectTemplatesLength = 0
	}
	objectTemplates := make([]map[string]interface{}, 0, objectTemplatesLength)
	policyTemplates := make([]map[string]map[string]interface{}, 0, policyTemplatesLength)
```

internal/plugin.go
```diff
@@ -209,22 +209,21 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 		p.PolicyDefaults.Controls = defaults.Controls
 	}
 
-	// Policy expanders default to true unless explicitly set in the config.
-	// Gatekeeper policy expander policyDefault
-	igvValue, set := getDefaultBool(unmarshaledConfig, "informGatekeeperPolicies")
```
mprahl, line 214: It looks like you accidentally deleted this code during your rebase. Could you please fix it?

internal/plugin.go
```diff
@@ -251,22 +250,21 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			policy.Controls = p.PolicyDefaults.Controls
 		}
 
-		// Policy expanders default to the policy default unless explicitly set.
-		// Gatekeeper policy expander policy override
-		igvValue, set := getPolicyBool(unmarshaledConfig, i, "informGatekeeperPolicies")
```
mprahl, line 256: Same as above

mprahl (approved):
Nice!

PR #14, opened 2021-09-16
Add support for Go v1.15

mprahl:
`os.ReadFile` is not available in Go 1.15, and this is required to build
on top of the UBI8 container image.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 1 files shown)


acm-grc-security (approved):

ChunxiAlexLuo:
> `os.ReadFile` is not available in Go 1.15, and this is required to build
> on top of the UBI8 container image.

Thanks for fixing this, just found the same issue

PR #15, opened 2021-09-16
Add support for automatically informing on Kyverno policy violations

mprahl:
By default, when a manifest includes a Kyverno policy, an additional
configuration policy will get generated to inform on Kyverno policy
violations in Open Cluster Management.

Additionally, remediation action is no longer set at the root policy level.
This allows the policy generator to have granular control of the
remediation action of configuration policies it may automatically
generator on behalf of the user (i.e. Kyverno policy reporting).

### DIFF
--- a/internal/expanders/expanders.go
+++ b/internal/expanders/expanders.go
@@ -1,0 +1,24 @@
+// Copyright Contributors to the Open Cluster Management project
+package expanders
+
+import (
+	"github.com/open-cluster-management/policy-generator-plugin/internal/types"
+)
+
+func GetExpanders() map[string]Expander {
+	return map[string]Expander{
+		"kyverno": KyvernoPolicyExpander{},
+	}
+}
+
+type Expander interface {
+	CanHandle(manifest map[string]interface{}) bool
+	Enabled(policyConf *types.PolicyConfig) bool
+	Expand(manifest map[string]interface{}, severity string) []map[string]map[string]interface{}
+}
+
+// Common constants for the expanders.
+const (
+	configPolicyAPIVersion = "policy.open-cluster-management.io/v1"
+	configPolicyKind       = "ConfigurationPolicy"
+)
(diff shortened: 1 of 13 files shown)


internal/expanders/expanders.go
```diff
@@ -0,0 +1,19 @@
...
+func GetExpanders() map[string]Expander {
+	return map[string]Expander{
+		"kyverno": KyvernPolicyExpander{},
+	}
+}
+
+type Expander interface {
+	CanHandle(manifest map[string]interface{}) bool
+	Expand(manifest map[string]interface{}, severity string) []map[string]map[string]interface{}
+}
```
dhaiducek, line 13: Ah, interface. Nice. Yes, this is what it should be. 🙂 

I was wondering: Do we want to add something like an `ExpandSetting` string to the interface and then use loops to iterate over each setting for the policyDefault object and policies array? In this case, it'd be "InformKyvernoPolicies" and then the logic for expansion can truly only be in the `expanders.go` file. Or is that going overboard?

internal/expanders/kyverno.go
```diff
@@ -0,0 +1,95 @@
+// Copyright Contributors to the Open Cluster Management project
+package expanders
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type KyvernPolicyExpander struct{}
```
dhaiducek, line 10: Minor nit: Missing an "o" in "Kyverno"

internal/expanders/expanders.go
dhaiducek: Okay, so the `Expander` would have to be a Struct to do that...

mprahl: It's a good idea. I'll add a method that checks if the expander is enabled for the policy.

dhaiducek:
LGTM!

PR #12, opened 2021-09-15
Gatekeeper policy generation

dhaiducek:

PR #16, opened 2021-09-20
Address comments from #12

mprahl:
This addresses the comments from #12 since Dale is out today and thus is not able to address the feedback.


ChunxiAlexLuo (approved):
/lgtm

PR #19, opened 2021-09-27
Build the release binaries statically

mprahl:
This resolves the issue on some systems that get the error:
PolicyGenerator: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by PolicyGenerator)

Signed-off-by: mprahl <mprahl@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.dylib
 **/.DS_Store
 PolicyGenerator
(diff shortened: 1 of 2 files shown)


ChunxiAlexLuo (approved):
/lgtm

looks good to me since we do not use any C language here

PR #24, opened 2021-10-13
Add the ability to override complianceType at the manifest level

mprahl:


internal/utils.go
```diff
@@ -120,42 +121,45 @@ func getManifests(policyConf *types.PolicyConfig) ([]map[string]interface{}, err
...
-			objectTemplates = append(objectTemplates, objTemplate)
-		} else {
-			// casting each objTemplate with manifest to objectTemplates type
-			// build policyTemplate for each objectTemplates
-			policyTemplate := buildPolicyTemplate(policyConf, &[]map[string]interface{}{objTemplate})
-			setNamespaceSelector(policyConf, policyTemplate)
-			policyTemplates = append(policyTemplates, *policyTemplate)
+	for i, manifestGroup := range manifestGroups {
+		complianceType := policyConf.Manifests[i].ComplianceType
+		for _, manifest := range manifestGroup {
```
serngawy, line 139: I believe we could flat this nested loop but lets keep it as enhancement for future work

PR #28, opened 2021-10-21
Fix duplicate names when a YAML file has multiple manifests

mprahl:
This only occurs with the consolidateManifests=false option.


internal/utils.go
```diff
@@ -148,7 +148,7 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 				// casting each objTemplate with manifest to objectTemplates type
 				// build policyTemplate for each objectTemplates
 				policyTemplate := buildPolicyTemplate(
-					policyConf, i+j+1, &[]map[string]interface{}{objTemplate},
+					policyConf, len(policyTemplates)+1, &[]map[string]interface{}{objTemplate},
```
ChunxiAlexLuo, line 151: Can we directly use `policyTemplatesLength + 1` here?

ChunxiAlexLuo (approved):
/lgtm

Went through with some test cases and logic looks good to me. Sorry for missing the identical index issue in yesterday PR review

PR #29, opened 2021-12-02
Restrict manifest paths to be in the kustomize directory tree

mprahl:
This is an added security measure so that the user cannot specify a
manifest path that is not related to their Kustomize configuration.

Relates:
https://github.com/open-cluster-management/backlog/issues/18236

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -9,6 +9,8 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
+	"strings"
 
 	"github.com/open-cluster-management/policy-generator-plugin/internal/expanders"
 	"github.com/open-cluster-management/policy-generator-plugin/internal/types"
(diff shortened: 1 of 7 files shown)


internal/utils.go
```diff
@@ -280,3 +282,30 @@ func unmarshalManifestBytes(manifestBytes []byte) (*[]map[string]interface{}, er
...
+	relPath, err := filepath.Rel(baseDirectory, absPath)
+	if err != nil {
+		return fmt.Errorf(
+			"could not resolve the manifest path %s to a relative path from the kustomization.yaml file",
+			manifestPath,
+		)
+	}
+
+	parDir := ".." + string(filepath.Separator)
+	if strings.HasPrefix(relPath, parDir) || relPath == ".." {
```
dhaiducek, line 303: Do you even need `relPath`? Couldn't this just be:
```go
if strings.HasPrefix(baseDirectory, absPath)
```

mprahl: @dhaiducek in most cases, I think this would work but your suggestion would not work if the base directory is `/var/path/to/base` and `absPath` is `/var/path/to/base-camp/symlink-to-something`. It's very unlikely that we would encounter this in app subscription, but I think it's better to use the current solution in the PR.

What do you think?

dhaiducek: Ah, that's a good point. Yes, that is safer.

dhaiducek (approved):

PR #35, opened 2022-02-02
Count policyset placement and validate config

ycao56:
https://github.com/stolostron/backlog/issues/19644
Policyset placement was not counted in config validation which was causing
placement to be generated by default

Signed-off-by: Yu Cao <ycao@redhat.com>

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -587,6 +587,76 @@ func (p *Plugin) assertValidConfig() error {
 		}
 	}
 
+	for i := range p.PolicySets {
+		seen := map[string]bool{}
+		plcset := &p.PolicySets[i]
+
+		if plcset.Name == "" {
+			return fmt.Errorf(
+				"each policySet must have a name set, but did not find a name at policySet array index %d", i,
+			)
+		}
+
+		if seen[plcset.Name] {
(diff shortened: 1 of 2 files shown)


gparvin (approved):
Thanks for adding the extra error checking for me!

internal/plugin.go
```diff
@@ -587,6 +587,76 @@ func (p *Plugin) assertValidConfig() error {
...
+		seen := map[string]bool{}
+		plcset := &p.PolicySets[i]
+
+		if plcset.Name == "" {
+			return fmt.Errorf(
+				"each policySet must have a name set, but did not find a name at policySet array index %d", i,
+			)
+		}
+
+		if seen[plcset.Name] {
```
mprahl, line 600: @ycao56 could you please declare a new set for `seen` in the event the policy set and policy name are the same?

ycao56: sure!

ycao56: `seen` is reassigned on line 591. So it should be safe?

mprahl: As discussed in chat, this needs to be defined outside of the loop.

PR #40, opened 2022-03-07
Change PolicySet API version to v1beta1

ycao56:
https://github.com/stolostron/backlog/issues/20475
Signed-off-by: Yu Cao <ycao@redhat.com>

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -18,10 +18,12 @@ import (
 
 const (
 	configPolicyKind           = "ConfigurationPolicy"
-	policyAPIVersion           = "policy.open-cluster-management.io/v1"
+	policyAPIGroup             = "policy.open-cluster-management.io"
+	policyAPIVersion           = policyAPIGroup + "/v1"
 	policyKind                 = "Policy"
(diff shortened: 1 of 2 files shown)


internal/plugin.go
```diff
@@ -18,7 +18,9 @@ import (
 
 const (
 	configPolicyKind           = "ConfigurationPolicy"
-	policyAPIVersion           = "policy.open-cluster-management.io/v1"
+	policyAPIGroup             = "policy.open-cluster-management.io"
+	v1                         = "v1"
+	v1beta1                    = "v1beta1"
```
dhaiducek, line 23: I'm not sure I see the benefit of naming the variable names the same as their content. Would it make more sense to call them `policyAPIVersion` and `policySetAPIVersion`? And then we can change the content without needing to update the variable name.

ycao56: This would save us two split calls on line 1147 and 1156

ycao56: Let me update. I can make it variable too

dhaiducek (approved):
That works! Sorry, I should have been clearer--I was thinking more like `policyAPIVersion=v1`, but this works too (and is probably even cleaner! 🙂

PR #41, opened 2022-03-16
Support policy type manifest 

ChunxiAlexLuo:
Signed-off-by: Chunxi Luo <chuluo@redhat.com>

https://github.com/stolostron/backlog/issues/19466

In the test cases I only used IAM examples, Certificate should be similar since they are directly used as policy-template and I can also add Certificate and other examples in the test cases if needed.

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -90,6 +90,12 @@ policies:
         # kustomization.yaml file. This path cannot be in a directory outside of the directory with
         # the kustomization.yaml file. Subdirectories within the directory with kustomization.yaml
         # file are allowed.
+        # Supported manifests:
+        #   1) Non-root policy type manifests such as IamPolicy, CertificatePolicy, and ConfigurationPolicy
+        #      that have a "Policy" suffix. These are not modified except for patches and are directly added 
+        #      as a Policy's policy-templates entry.
+        #   2) For everything else, ConfigurationPolicy objects are generated to wrap these manifests. 
+        #      The resulting ConfigurationPolicy is added as a Policy's policy-templates entry.
       - path: ""
         # Optional. (See policy[0].complianceType for description.)
         complianceType: "musthave"
--- a/internal/plugin_config_test.go
+++ b/internal/plugin_config_test.go
@@ -28,6 +28,28 @@ data:
 	}
 }
 
+func createIamPolicyManifest(t *testing.T, tmpDir, filename string) {
+	t.Helper()
+	manifestsPath := path.Join(tmpDir, filename)
+	yamlContent := `
+apiVersion: policy.open-cluster-management.io/v1
+kind: IamPolicy
+metadata:
+  name: policy-limitclusteradmin-example
+spec:
+  severity: medium
+  namespaceSelector:
+    include: ["*"]
+    exclude: ["kube-*", "openshift-*"]
+  remediationAction: enforce
+  maxClusterRoleBindingUsers: 5
+`
+	err := ioutil.WriteFile(manifestsPath, []byte(yamlContent), 0o666)
+	if err != nil {
+		t.Fatalf("Failed to write %s", manifestsPath)
+	}
+}
+
 func TestConfig(t *testing.T) {
 	t.Parallel()
 	tmpDir := t.TempDir()
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -15,6 +15,7 @@ import (
 	"github.com/stolostron/policy-generator-plugin/internal/expanders"
 	"github.com/stolostron/policy-generator-plugin/internal/types"
 	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 // getManifests will get all of the manifest files associated with the input policy configuration
@@ -139,12 +140,28 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	for i, manifestGroup := range manifestGroups {
 		complianceType := policyConf.Manifests[i].ComplianceType
 		for _, manifest := range manifestGroup {
+			isPolicyTypeManifest, err := isPolicyTypeManifest(manifest)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"%w in manifest path: %s",
+					err,
+					policyConf.Manifests[i].Path,
+				)
+			}
+
+			if isPolicyTypeManifest {
+				policyTemplate := map[string]map[string]interface{}{"objectDefinition": manifest}
+				policyTemplates = append(policyTemplates, policyTemplate)
+
+				continue
+			}
+
 			objTemplate := map[string]interface{}{
 				"complianceType":   complianceType,
 				"objectDefinition": manifest,
 			}
 			if policyConf.ConsolidateManifests {
-				// put all objTemplate with manifest into single consolidated objectTemplates object
+				// put all objTemplate with manifest into single consolidated objectTemplates
 				objectTemplates = append(objectTemplates, objTemplate)
 			} else {
 				// casting each objTemplate with manifest to objectTemplates type
@@ -167,8 +184,9 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 		)
 	}
 
-	//  just build one policyTemplate by using the above consolidated objectTemplates
-	if policyConf.ConsolidateManifests {
+	// just build one policyTemplate by using the above non-empty consolidated objectTemplates
+	// ConsolidateManifests = true or there is non-policy-type manifest
+	if policyConf.ConsolidateManifests && len(objectTemplates) > 0 {
 		policyTemplate := buildPolicyTemplate(policyConf, 1, &objectTemplates, &policyConf.EvaluationInterval)
 		setNamespaceSelector(policyConf, policyTemplate)
 		policyTemplates = append(policyTemplates, *policyTemplate)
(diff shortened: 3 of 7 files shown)


internal/utils.go
```diff
@@ -183,6 +191,26 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	return policyTemplates, nil
 }
 
+// checkPolicyTypeManifest determine if the manifest is policy-type manifest.
+func checkPolicyTypeManifest(manifest map[string]interface{}) bool {
+	apiVersion, foundAPIVersion, isStr := unstructured.NestedString(manifest, "apiVersion")
+	if !foundAPIVersion || isStr != nil {
+		return false
+	} else if strings.HasPrefix(apiVersion, "policy.open-cluster-management.io") {
```
dhaiducek, line 199: This has the potential to nest a `Policy`, `PolicySet`, etc. kind inside of a policy, which I'm not sure we'd want. Should we check against a list of supported kinds?

@mprahl @ycao56 What do you think?

dhaiducek (dismissed):
Would you please also add an example to the `examples` directory and add to the comments in the `policygenerator-reference.yaml`?

ChunxiAlexLuo:
> Would you please also add an example to the `examples` directory and add to the comments in the `policygenerator-reference.yaml`?

sure, since no change in policygenerator-reference.yaml, I will add the comments under manifest.path section

internal/utils.go
```diff
@@ -139,24 +140,30 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
-				policyTemplate := buildPolicyTemplate(
-					policyConf,
-					len(policyTemplates)+1,
-					&[]map[string]interface{}{objTemplate},
-					&policyConf.Manifests[i].EvaluationInterval,
-				)
-				setNamespaceSelector(policyConf, policyTemplate)
-				policyTemplates = append(policyTemplates, *policyTemplate)
+			isPolicyTypeManifest := checkPolicyTypeManifest(manifest)
+			if !isPolicyTypeManifest { // build objTemplate from non-policy-type manifest
```
mprahl, line 144: I find it'd be easier to read if you could do the following:
```go
if isPolicyTypeManifest {
	policyTemplate := buildFromPolicyTypeManifest(manifest)
	policyTemplates = append(policyTemplates, policyTemplate)

	continue
}

objTemplate := map[string]interface{}{
...
```

internal/utils.go
mprahl: I would also like the `Kind` to be checked to see if the string ends with `Policy` and also doesn't equal `Policy`.

internal/utils.go
```diff
@@ -183,6 +191,26 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	return policyTemplates, nil
 }
 
+// checkPolicyTypeManifest determine if the manifest is policy-type manifest.
```
mprahl, line 194:
```suggestion
// checkPolicyTypeManifest determines if the manifest is a non-root policy manifest.
```

internal/utils.go
```diff
@@ -183,6 +191,26 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
+func checkPolicyTypeManifest(manifest map[string]interface{}) bool {
```
mprahl, line 195: This is a little easier to understand in my opinion.
```suggestion
func isPolicyTypeManifest(manifest map[string]interface{}) bool {
```

internal/utils.go
```diff
@@ -183,6 +191,26 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
+	apiVersion, foundAPIVersion, isStr := unstructured.NestedString(manifest, "apiVersion")
+	if !foundAPIVersion || isStr != nil {
```
mprahl, line 197: I think you can remove this if statement since in the error case, `apiVersion` is an empty string so the `HasPrefix` below would still be false.

internal/utils.go
```diff
@@ -183,6 +191,26 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
...
+	} else if strings.HasPrefix(apiVersion, "policy.open-cluster-management.io") {
+		return true // policy-type manifest contains policy api
+	}
+
+	return false
+}
+
+// buildFromPolicyTypeManifest take non-root policy-type manifest
+// and directly return as a policyTemplate.
+func buildFromPolicyTypeManifest(
```
mprahl, line 208: Since this code is only used once, I think you could get rid of this function and just use the code directly.

internal/plugin_config_test.go
```diff
@@ -28,6 +28,28 @@ data:
 	}
 }
 
+func createIamPolicyTypeConfigMap(t *testing.T, tmpDir, filename string) {
```
mprahl, line 31: This is clearer in my opinion.
```suggestion
func createIamPolicyManifest(t *testing.T, tmpDir, filename string) {
```

mprahl (changes requested):

internal/utils.go
dhaiducek: That's a good idea--then we wouldn't have to maintain a list of policy kinds.

internal/utils.go
ChunxiAlexLuo: I check these in case apiVersion or kind are objects rather than string then HasPrefix will throw some error.

ycao56:
@ChunxiAlexLuo can you update the commit message to follow the guideline we agreed on previously?

docs/policygenerator-reference.yaml
```diff
@@ -90,6 +90,12 @@ policies:
         # kustomization.yaml file. This path cannot be in a directory outside of the directory with
         # the kustomization.yaml file. Subdirectories within the directory with kustomization.yaml
         # file are allowed.
+        # Manifests format:
+        # 1) For policy-type manifests under manifests.path, which are already non-root policies (IamPolicy,
```
ycao56, line 94: Suggest to rephrase to:
Supported manifests:
1) Policy type manifests such as IamPolicy, CertificatePolicy and ConfigurationPolicy

docs/policygenerator-reference.yaml
```diff
@@ -90,6 +90,11 @@ policies:
         # kustomization.yaml file. This path cannot be in a directory outside of the directory with
         # the kustomization.yaml file. Subdirectories within the directory with kustomization.yaml
         # file are allowed.
+        # Supported manifests:
+        # 1) Policy type manifests such as IamPolicy, CertificatePolicy and ConfigurationPolicy,etc.
+        # The generator treat them as policy-template.
+        # 2) Other kinds of manifests. The generator will use them as object-template
+        # component to building ConfigurationPolicy kind policy-template
```
mprahl, line 97:
```suggestion
        # Supported manifests:
        # 1) Policy type manifests such as IamPolicy, CertificatePolicy, ConfigurationPolicy, and etc.
        # These are not modified except for patches and are directly added as a Policy's policy-templates entry.
        # 2) For everything else, ConfigurationPolicy objects are generated to wrap these manifests. The resulting ConfigurationPolicy is added as a Policy's policy-templates entry.
```

internal/utils_test.go
```diff
@@ -233,6 +233,123 @@ data:
...
+	err := ioutil.WriteFile(IamManifestPath, []byte(IamManifestYAML), 0o666)
+	if err != nil {
+		t.Fatalf("Failed to write %s", IamManifestPath)
+	}
+
+	manifestFiles = append(
+		manifestFiles, types.Manifest{Path: IamManifestPath},
+	)
+
+	policysetManifestPath := path.Join(tmpDir, "configmapPolicyset.yaml")
```
mprahl, line 266: Could you also not use `configmap` in the file name? It's confusing since it implies the file is a `ConfigMap` object.

internal/utils_test.go
```diff
@@ -233,6 +233,123 @@ data:
...
+		t.Fatalf("Failed to write %s", policysetManifestPath)
+	}
+
+	manifestFiles = append(
+		manifestFiles, types.Manifest{Path: policysetManifestPath},
+	)
+
+	// Write a bogus file to ensure it is not picked up when creating the policy
+	// template
+	bogusFilePath := path.Join(tmpDir, "README.md")
```
mprahl, line 292: This is already tested else where, so could you please remove this to make the tests "slimmer"?

mprahl (approved):
/hold for other reviewers

mprahl (changes requested):
Sorry, one last comment I forgot about.

dhaiducek (commented):
Great work, @ChunxiAlexLuo!

I've left some review comments for consideration. I also noticed there are no unit tests for `isPolicyTypeManifest`. Unless I missed it, could we add tests there?

ycao56 (dismissed):

ycao56 (approved):

internal/utils_test.go
```diff
@@ -233,6 +233,140 @@ data:
 	}
 }
 
+func TestIsPolicyTypeManifest(t *testing.T) {
+	t.Parallel()
+
+	manifestIam := map[string]interface{}{
```
mprahl, line 239: Could you please break these up into subtests like so (untested:
```go
tests := []struct{
    apiVersion string
    kind string
    expected bool
}{
    {"policy.open-cluster-management.io/v1", "IamPolicy},
    ...
}

for _, test := range tests {
    test := test
    t.Run(
        fmt.Sprintf("apiVersion=%s, kind=%s", test.apiVersion, test.kind),
        func(t *testing.T) {
            t.Parallel()
            manifest := map[string]interface{}{
                "apiVersion": test.apiVersion,
                "kind": test.kind,
            }
            isPolicyType, err = isPolicyTypeManifest(manifest)
            assertEqual(t, isPolicyType, test.expected)
            assertEqual(t, err, nil)
        }
    )
}
```

internal/utils.go
```diff
@@ -139,12 +140,24 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	for i, manifestGroup := range manifestGroups {
 		complianceType := policyConf.Manifests[i].ComplianceType
 		for _, manifest := range manifestGroup {
+			isPolicyTypeManifest, err := isPolicyTypeManifest(manifest)
+			if err != nil {
+				return nil, err
```
mprahl, line 145: Could you add additional context to the error so the user knows which manifest has the invalid `apiVersion` or `kind`?

internal/utils.go
```diff
@@ -139,12 +140,28 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	for i, manifestGroup := range manifestGroups {
 		complianceType := policyConf.Manifests[i].ComplianceType
 		for _, manifest := range manifestGroup {
+			isPolicyTypeManifest, err := isPolicyTypeManifest(manifest)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"%s in manifest path: %s",
+					err.Error(),
```
mprahl, line 147:
```suggestion
					"%w in manifest path: %s",
					err,
```

mprahl (approved):
Nice!

PR #42, opened 2022-03-22
add ability to build c shared-library for `processGeneratorConfig`

TheRealHaoLiu:
- add C type compatible wrapper function for `processGeneratorConfig`
- add `build-so` Make target for generating C share-object file


Makefile
```diff
@@ -36,6 +38,14 @@ build: layout
...
 	go build -o PolicyGenerator cmd/main.go
 
+build-so:
+	@mkdir -p build_output
+		go build -buildmode=c-shared -o build_output/_processGeneratorConfigC_$(OS)_$(ARCH).so cmd/main.go
+
+build-so-multiarch:
+	@mkdir -p build_output
+	GOOS=linux CGO_ENABLED=1 GOARCH=amd64 go build -o build_output/_processGeneratorConfigC_linux_amd64.so cmd/main.go
+
```
dhaiducek, line 48: This Make target throws all sorts of errors on my MacBook, but is this Linux-specific? The "multiarch" name is throwing me off a bit.

dhaiducek:
Thanks for the PR! What's the context here for the C wrapper?

cmd/main.go
```diff
@@ -74,3 +75,13 @@ func processGeneratorConfig(filePath string) []byte {
 
 	return generatedOutput
 }
+
+// C compatible function for processGeneratorConfig
+// Enable C shared-library to be compiled and use by Python
```
mprahl, line 80:
```suggestion
// Enable C shared-library to be compiled and used by Python
```

Makefile
```diff
@@ -36,6 +38,14 @@ build: layout
 build-binary:
 	go build -o PolicyGenerator cmd/main.go
 
+build-so:
+	@mkdir -p build_output
+		go build -buildmode=c-shared -o build_output/_processGeneratorConfigC_$(OS)_$(ARCH).so cmd/main.go
```
dhaiducek, line 43:
```suggestion
build-so:
	@mkdir -p build_output
	go build -buildmode=c-shared -o build_output/_processGeneratorConfigC_$(OS)_$(ARCH).so cmd/main.go
```

Makefile
```diff
@@ -36,6 +38,14 @@ build: layout
...
+
+build-so-multiarch:
```
mprahl, line 45: Could you explain why this is `multiarch`? It seems that it's just producing an `amd64` SO.

mprahl:
@TheRealHaoLiu will you need the shared object included as an artifact on GitHub releases?

Makefile
TheRealHaoLiu: yah i forgot to add other archs....also cross compile CGO on mac... is hard

i can remove this target for now

mprahl:
> @mprahl does the github release process able to build multiarch? how would i be able to try that?

Does compiling the shared object respect `GOARCH`?

PR #43, opened 2022-03-25
Add-support-for-the-metadataComplianceType-field

ChunxiAlexLuo:
https://github.com/stolostron/backlog/issues/19517

Signed-off-by: Chunxi Luo <chuluo@redhat.com>

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -37,6 +37,9 @@ policyDefaults:
   # configuration policy should be generated in order to receive policy violations in Open Cluster
   # Management when the Kyverno policy has been violated. This defaults to true.
   informKyvernoPolicies: true
+  # Optional. Overrides complianceType when comparing the manifest's metadata section to objects on
+  # the cluster ("musthave",  "mustonlyhave"). Default is unset to not override complianceType for metadata.
+  metadataComplianceType: ""
   # Required. The namespace of all the policies.
   namespace: ""
   # Optional. The placement configuration for the policies. This defaults to a placement
@@ -99,6 +102,8 @@ policies:
       - path: ""
         # Optional. (See policy[0].complianceType for description.)
         complianceType: "musthave"
+        # Optional. (See policyDefaults.metadataComplianceType for description.)
+        metadataComplianceType: ""
          # Optional. (See policyDefaults.evaluationInterval for description.)
         evaluationInterval: {}
         # Optional. A Kustomize patch to apply to the manifest(s) at the path. If there
@@ -127,6 +132,8 @@ policies:
     # Optional. Determines the policy controller behavior when comparing the manifest to objects on
     # the cluster ("musthave",  "mustonlyhave", or "mustnothave"). Defaults to "musthave".
     complianceType: "musthave"
+    # Optional. (See policyDefaults.metadataComplianceType for description.)
+    metadataComplianceType: ""
     # Optional. (See policyDefaults.controls for description.)
     controls:
       - "CM-2 Baseline Configuration"
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -394,6 +394,10 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			policy.ComplianceType = p.PolicyDefaults.ComplianceType
 		}
 
+		if policy.MetadataComplianceType == "" && p.PolicyDefaults.MetadataComplianceType != "" {
+			policy.MetadataComplianceType = p.PolicyDefaults.MetadataComplianceType
+		}
+
 		if policy.Controls == nil {
 			policy.Controls = p.PolicyDefaults.Controls
 		}
@@ -494,6 +498,10 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 
 			if manifest.ComplianceType == "" {
 				manifest.ComplianceType = policy.ComplianceType
+			}
+
+			if manifest.MetadataComplianceType == "" && policy.MetadataComplianceType != "" {
+				manifest.MetadataComplianceType = policy.MetadataComplianceType
 			}
 
 			// If the manifests are consolidated to a single ConfigurationPolicy object, don't set
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -139,6 +139,7 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 	policyTemplates := make([]map[string]map[string]interface{}, 0, policyTemplatesLength)
 	for i, manifestGroup := range manifestGroups {
 		complianceType := policyConf.Manifests[i].ComplianceType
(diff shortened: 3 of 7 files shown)


internal/plugin.go
```diff
@@ -61,12 +61,13 @@ type Plugin struct {
...
 var defaults = types.PolicyDefaults{
-	Categories:        []string{"CM Configuration Management"},
-	ComplianceType:    "musthave",
-	Controls:          []string{"CM-2 Baseline Configuration"},
-	RemediationAction: "inform",
-	Severity:          "low",
-	Standards:         []string{"NIST SP 800-53"},
+	Categories:             []string{"CM Configuration Management"},
+	ComplianceType:         "musthave",
+	MetadataComplianceType: "",
```
ChunxiAlexLuo, line 66: Here based the logic here https://github.com/stolostron/config-policy-controller/blob/main/controllers/configurationpolicy_controller.go#L1845, default MetadataComplianceType need to be set as `""` rather than `"musthave"`

internal/expanders/kyverno.go
```diff
@@ -64,7 +64,8 @@ func (k KyvernoPolicyExpander) Expand(
 				"severity":          severity,
 				"object-templates": []map[string]interface{}{
 					{
-						"complianceType": "mustnothave",
+						"complianceType":         "mustnothave",
+						"metadataComplianceType": "",
```
ChunxiAlexLuo, line 68: @dhaiducek  @mprahl Need to confirm metadataComplianceType in kyverno expander

internal/expanders/gatekeeper.go
```diff
@@ -68,7 +68,8 @@ func (g GatekeeperPolicyExpander) Expand(
 				"severity":          severity,
 				"object-templates": []map[string]interface{}{
 					{
-						"complianceType": "musthave",
+						"complianceType":         "musthave",
+						"metadataComplianceType": "",
```
ChunxiAlexLuo, line 72: @dhaiducek  @mprahl Need to confirm metadataComplianceType in gatekeeper expander

docs/policygenerator-reference.yaml
```diff
@@ -127,6 +129,9 @@ policies:
     # Optional. Determines the policy controller behavior when comparing the manifest to objects on
     # the cluster ("musthave",  "mustonlyhave", or "mustnothave"). Defaults to "musthave".
     complianceType: "musthave"
+    # Optional. Overriding complianceType when comparing the manifest's metadata section to objects on
+    # the cluster ("musthave",  "mustonlyhave"). Defaults is unset and still using complianceType.
+    metadataComplianceType: ""
```
dhaiducek, line 134:
```suggestion
    # Optional. Override complianceType when comparing the manifest's metadata section to objects on
    # the cluster ("musthave",  "mustonlyhave"). Default is unset to not override complianceType for metadata.
    metadataComplianceType: ""
```

internal/plugin.go
```diff
@@ -394,6 +395,8 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			policy.ComplianceType = p.PolicyDefaults.ComplianceType
 		}
 
+		// No need to set policy.MetadataComplianceType here since defaults value is ""
```
dhaiducek, line 398:
```suggestion
		// No need to set policy.MetadataComplianceType here since default value is ""
```

internal/expanders/gatekeeper.go
mprahl: @ChunxiAlexLuo I would leave this out in the expanders.

internal/plugin.go
mprahl: This defaults to an empty string in Go, so this line is unnecessary.

docs/policygenerator-reference.yaml
```diff
@@ -127,6 +129,9 @@ policies:
     # Optional. Determines the policy controller behavior when comparing the manifest to objects on
     # the cluster ("musthave",  "mustonlyhave", or "mustnothave"). Defaults to "musthave".
     complianceType: "musthave"
+    # Optional. Overrides complianceType when comparing the manifest's metadata section to objects on
+    # the cluster ("musthave",  "mustonlyhave"). Default is unset to not override complianceType for metadata.
+    metadataComplianceType: ""
```
mprahl, line 134: Could you please also document this under `policyDefaults`?

internal/plugin.go
```diff
@@ -394,6 +395,8 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			policy.ComplianceType = p.PolicyDefaults.ComplianceType
 		}
 
+		// No need to set policy.MetadataComplianceType here since default value is ""
```
mprahl, line 398: This isn't quite right since the user can set a default to this value, for example:
```
policyDefaults:
  metadataComplianceType: musthave
```

internal/utils.go
```diff
@@ -157,8 +158,9 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]map[string
 			}
 
 			objTemplate := map[string]interface{}{
-				"complianceType":   complianceType,
-				"objectDefinition": manifest,
+				"complianceType":         complianceType,
+				"metadataComplianceType": metadataComplianceType,
```
mprahl, line 162: Could you only set this if `metadataComplianceType != ""`? That way the generated policy is a little cleaner when this value isn't set.

mprahl (changes requested):

docs/policygenerator-reference.yaml
ChunxiAlexLuo: I did it as current PR  because I'm not sure if we also want `metadataComplianceType` under `policyDefaults`, or just under `policies` and `manifests` as `complianceType`, since `metadataComplianceType` is an enhancement of `complianceType`. If confirmed, I can update it here.

internal/plugin.go
ChunxiAlexLuo: I did it as current PR  because I'm not sure if we also want `metadataComplianceType` under `policyDefaults`, or just under `policies` and `manifests` as `complianceType`, since `metadataComplianceType` is an enhancement of `complianceType`. If confirmed, I can update it here.

mprahl: @ChunxiAlexLuo yes, but you added it to the `PolicyDefaults` structure, which makes it available under `policyDefaults` in the YAML.

mprahl: Also, just about every other option allows for inheritance from `policyDefaults` -> `policies`, so it'd be nice to be consistent here as well.

ChunxiAlexLuo:
@mprahl  So two updates: 1) add `metadataComplianceType` under `policyDefault`, 2) only add non empty `metadataComplianceType` into `objTemplate`, updated all test cases for these

mprahl (commented):
One small change and then this is good to merge. Nice job!

mprahl (approved):
Nice!

PR #44, opened 2022-04-02
Allow adding policy annotations

serngawy:
Signed-off-by: melserngawy <melserng@redhat.com>

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -386,8 +386,25 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 
 	for i := range p.Policies {
 		policy := &p.Policies[i]
+
+		if policy.PolicyAnnotations == nil {
+			annotations := map[string]string{}
+			for k, v := range p.PolicyDefaults.PolicyAnnotations {
+				annotations[k] = v
+			}
+			policy.PolicyAnnotations = annotations
+		}
+
 		if policy.Categories == nil {
 			policy.Categories = p.PolicyDefaults.Categories
+		}
+
+		if policy.Standards == nil {
+			policy.Standards = p.PolicyDefaults.Standards
+		}
+
+		if policy.Controls == nil {
+			policy.Controls = p.PolicyDefaults.Controls
 		}
 
 		if policy.ComplianceType == "" {
@@ -396,10 +413,6 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 
 		if policy.MetadataComplianceType == "" && p.PolicyDefaults.MetadataComplianceType != "" {
 			policy.MetadataComplianceType = p.PolicyDefaults.MetadataComplianceType
-		}
-
-		if policy.Controls == nil {
-			policy.Controls = p.PolicyDefaults.Controls
 		}
 
 		// Only use the policyDefault evaluationInterval value when it's not explicitly set on the policy.
@@ -487,10 +500,6 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 
 		if policy.Severity == "" {
 			policy.Severity = p.PolicyDefaults.Severity
-		}
-
-		if policy.Standards == nil {
-			policy.Standards = p.PolicyDefaults.Standards
 		}
 
 		for j := range policy.Manifests {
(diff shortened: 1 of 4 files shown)


internal/plugin.go
```diff
@@ -844,17 +844,27 @@ func (p *Plugin) createPolicy(policyConf *types.PolicyConfig) error {
 		return err
 	}
 
+	annotations := map[string]string{
+		"policy.open-cluster-management.io/categories": strings.Join(policyConf.Categories, ","),
+		"policy.open-cluster-management.io/controls":   strings.Join(policyConf.Controls, ","),
+		"policy.open-cluster-management.io/standards":  strings.Join(policyConf.Standards, ","),
+	}
+
+	for k, v := range p.PolicyDefaults.PolicyAnnotations {
```
mprahl, line 853: To match the rest of the code, could you apply the defaults on the policy configuration in the `applyDefaults` method?

serngawy: done

internal/plugin.go
```diff
@@ -329,6 +329,17 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 		p.PolicyDefaults.Controls = defaults.Controls
 	}
 
+	if len(p.PolicyDefaults.PolicyAnnotations) > 0 {
+		annotation := p.PolicyDefaults.PolicyAnnotations
+		for i := range p.Policies {
```
mprahl, line 334: Could you please reuse the `for i := range p.Policies {` loop below?

Also, like I said, could you only apply `p.PolicyDefaults.PolicyAnnotations` to the policy if the policy doesn't define a map. It shouldn't be a union. Also, if the policy has explicitly set `PolicyAnnotations` to an empty map, it shouldn't inherit the defaults.

serngawy: As a user when I define  `p.PolicyDefaults.PolicyAnnotations` that is mean I need to have this annotations in all generated policies from this policygenerator template. 
If there are some annotations I don't want them in all generated policies I would not add them in `p.PolicyDefaults.PolicyAnnotations` I add them to the desired `PolicyConfig.PolicyAnnotations` is that make sense ? do I miss understand ?

mprahl: I see your point, but after discussing with the team, I don't want the override/inheritance behavior to be different for this field compared to the other policy defaults.

mprahl:
Could you please document this new field in https://github.com/stolostron/policy-generator-plugin/blob/main/docs/policygenerator-reference.yaml?

internal/plugin.go
```diff
@@ -844,17 +848,23 @@ func (p *Plugin) createPolicy(policyConf *types.PolicyConfig) error {
...
 	}
 
+	annotations := map[string]string{
+		"policy.open-cluster-management.io/categories": strings.Join(policyConf.Categories, ","),
+		"policy.open-cluster-management.io/controls":   strings.Join(policyConf.Controls, ","),
+		"policy.open-cluster-management.io/standards":  strings.Join(policyConf.Standards, ","),
+	}
+
+	for k, v := range policyConf.PolicyAnnotations {
+		annotations[k] = v
```
mprahl, line 858: If a policy uses the default `PolicyAnnotations`, could you please create a copy of that in the `applyDefaults` method? That way, the implementation of `applyDefaults` does not need to be accounted for in `createPolicy`.

serngawy: its not what you mean, its already copy the DefaultPolicyAnnotation into the PolicyAnnotation 
https://github.com/stolostron/policy-generator-plugin/pull/44/files#diff-6e087f8d6059d270802616bc8a93ff47320e6302233e7ca9cf317b3c7054c8abR389

mprahl: @serngawy that copies the references but not the map itself. See this as an example:
https://riptutorial.com/go/example/9834/copy-a-map

serngawy: done

internal/plugin.go
```diff
@@ -386,10 +386,26 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 
 	for i := range p.Policies {
 		policy := &p.Policies[i]
+		if len(policy.PolicyAnnotations) == 0 {
```
mprahl, line 389: The issue with this approach is that I can't overwrite the policy default's annotations when setting an empty map at the policy level.

I'm not 100% sure, but I think if you change this to `if policy.PolicyAnnotations == nil {` it would work. Could you please try that?

internal/plugin.go
```diff
@@ -386,10 +386,26 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
...
+				annotations[k] = v
+			}
+			policy.PolicyAnnotations = annotations
+		}
+
 		if policy.Categories == nil {
 			policy.Categories = p.PolicyDefaults.Categories
 		}
 
+		if policy.Standards == nil {
```
mprahl, line 401: Optional: This change is unnecessary and it'd be better if this and the controls section were left like they were originally in order to keep the Git history clean.

internal/plugin_test.go
```diff
@@ -426,6 +426,111 @@ spec:
...
+	p.PolicyDefaults.Namespace = "my-policies"
+	p.PolicyDefaults.PolicyAnnotations = map[string]string{"test-default-annotation": "default"}
+
+	policyConf := types.PolicyConfig{
+		Name: "policy-app-config",
+		Manifests: []types.Manifest{
+			{Path: path.Join(tmpDir, "configmap.yaml")},
+		},
+	}
+	p.Policies = append(p.Policies, policyConf)
```
mprahl, line 443: Could you please add two more policies? The first one would override the default annotations with some other annotations. The other would override the default annotations with an empty map.

serngawy: add test case for empty map override.

mprahl (approved):
Nice job! Thanks for the contribution.

PR #48, opened 2022-05-11
Add support for policyDefaults.disabled

willkutler:

https://github.com/stolostron/backlog/issues/22350

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -25,6 +25,9 @@ policyDefaults:
   # manifests being wrapped in the policy. If set to false, a configuration policy per manifest will
   # be generated. This defaults to true.
   consolidateManifests: true
+  # Optional. Determines whether the policy is enabled or disabled. A disabled policy will not be
+  # propagated to any managed clusters and will show no status as a result.
+  disabled: false
   # Optional. This is how often a policy should be evaluated when in a particular compliance state.
   # When managed clusters have low CPU resources, the evaluation interval can be increased to
   # to reduce CPU usage on the Kubernetes API.
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -465,6 +465,13 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			policy.ConsolidateManifests = p.PolicyDefaults.ConsolidateManifests
 		}
 
+		disabledValue, setDisabled := getPolicyBool(unmarshaledConfig, i, "disabled")
+		if setDisabled {
+			policy.Disabled = disabledValue
+		} else {
+			policy.Disabled = p.PolicyDefaults.Disabled
+		}
+
 		// Determine whether defaults are set for placement
 		plcDefaultSet := len(p.PolicyDefaults.Placement.LabelSelector) != 0 || p.PolicyDefaults.Placement.PlacementPath != ""
 		plrDefaultSet := len(p.PolicyDefaults.Placement.ClusterSelectors) != 0 || p.PolicyDefaults.Placement.PlacementRulePath != ""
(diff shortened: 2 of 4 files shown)


docs/policygenerator-reference.yaml
```diff
@@ -139,7 +142,7 @@ policies:
     # Optional. (See policyDefaults.controls for description.)
     controls:
       - "CM-2 Baseline Configuration"
-    # Optional. (See policyDefaults.disabled for description.)
+    # Required. (See policyDefaults.disabled for description.)
```
dhaiducek, line 145:
```suggestion
    # Optional. (See policyDefaults.disabled for description.)
```

docs/policygenerator-reference.yaml
```diff
@@ -25,6 +25,9 @@ policyDefaults:
   # manifests being wrapped in the policy. If set to false, a configuration policy per manifest will
   # be generated. This defaults to true.
   consolidateManifests: true
+  # Required. Determines whether the policy is enabled or disabled. A disabled policy will not be
```
dhaiducek, line 28:
```suggestion
  # Optional. Determines whether the policy is enabled or disabled. A disabled policy will not be
```

dhaiducek (changes requested):
Looks good for the reference doc update, but for the purposes of the generator this value is optional (since it'll fall back to a default setting).

However, the code needs to be updated as well. It's not currently exposed in the `policyDefault` struct: https://github.com/stolostron/policy-generator-plugin/blob/a26a18edfef92b6127e9d0c355b62bc0eb814a1d/internal/types/types.go#L56

And then will need to be validated along with the other values: https://github.com/stolostron/policy-generator-plugin/blob/a26a18edfef92b6127e9d0c355b62bc0eb814a1d/internal/plugin.go#L314

mprahl:
@willkutler actually, the `disabled` isn't available under `policyDefaults`.

To add it, you'll need to do the following:
* Add the `Disabled` field to this struct https://github.com/stolostron/policy-generator-plugin/blob/a26a18edfef92b6127e9d0c355b62bc0eb814a1d/internal/types/types.go#L56
* Apply the `policyDefault.disabled` value on each policy that doesn't set it in the `applyDefaults` method: https://github.com/stolostron/policy-generator-plugin/blob/a26a18edfef92b6127e9d0c355b62bc0eb814a1d/internal/plugin.go#L314
* Use the helper function `getPolicyBool` to ensure that an explicit `false` wasn't set on the `policy.Disabled` before setting the default.

internal/plugin.go
```diff
@@ -352,6 +352,13 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 		p.PolicyDefaults.ConsolidateManifests = true
 	}
 
+	disabled, setDisabled := getDefaultBool(unmarshaledConfig, "disabled")
```
mprahl, line 355: This section isn't needed because the default value is `false` and the unmarshalled value of `p.PolicyDefaults.Disabled` will default to `false`, so we can just rely on that here.

This code would only be required if we default something to `true` but we want to detect if the user just omitted the value or actually explicitly set it to `false`.

mprahl (changes requested):
Could you please also add a unit test?

internal/plugin_config_test.go
```diff
@@ -1127,3 +1129,35 @@ func TestPolicySetConfig(t *testing.T) {
...
+		`
+apiVersion: policy.open-cluster-management.io/v1
+kind: PolicyGenerator
+metadata:
+  name: policy-generator-name
+policyDefaults:
+  namespace: my-policies
+  disabled: true
+policies:
+- name: policy-app-config
```
mprahl, line 1148: Could you please add another policy that overrides the `disabled` field from the defaults? Then just check to make sure the override is respected.

willkutler: @mprahl added one

mprahl:
@willkutler this looks great! Could you please squash your commits and change the commit message to something like "Add support for policyDefaults.disabled".

willkutler:
@mprahl done! thanks for the reviews

mprahl (approved):
Nice!

PR #62, opened 2022-08-18
Add LabelSelector to NamespaceSelector

dhaiducek:
I wrote a grotesquely monstrous custom NamespaceSelector comparison tool for the tests. Not sure if there was a better way to do that?

Addresses:
- https://github.com/stolostron/backlog/issues/25135

### DIFF
--- a/internal/utils_test.go
+++ b/internal/utils_test.go
@@ -30,6 +30,83 @@ func assertReflectEqual(t *testing.T, a interface{}, b interface{}) {
 	}
 }
 
+func assertSelectorEqual(t *testing.T, a map[string]interface{}, b types.NamespaceSelector) {
+	t.Helper()
+
+	if !compareSelectors(a, b) {
+		t.Fatalf("%s != %s", a, b)
+	}
+}
+
+func compareStringArrays(a []interface{}, b []string) bool {
+	// Account for when b is []string(nil)
+	if len(a) == 0 && len(b) == 0 {
+		return true
+	}
+
+	// Create a string array from []interface{}
+	aTyped := make([]string, len(a))
+	for i, val := range a {
+		aTyped[i] = val.(string)
+	}
+
+	return reflect.DeepEqual(aTyped, b)
+}
+
+func compareSelectors(a map[string]interface{}, b types.NamespaceSelector) bool {
+	if includeA, ok := a["include"].([]interface{}); ok {
+		if !compareStringArrays(includeA, b.Include) {
+			return false
+		}
+	} else if len(b.Include) != 0 {
+		return false
+	}
+
+	if excludeA, ok := a["exclude"].([]interface{}); ok {
+		if !compareStringArrays(excludeA, b.Exclude) {
+			return false
+		}
+	} else if len(b.Exclude) != 0 {
+		return false
+	}
+
+	if matchLabelsA, ok := a["matchLabels"].(map[string]string); ok {
+		if !reflect.DeepEqual(matchLabelsA, b.MatchLabels) {
+			return false
+		}
+	} else if matchLabelsA != nil && b.MatchLabels != nil {
+		return false
+	}
+
+	if matchExpressionsA, ok := a["matchExpressions"].([]interface{}); ok {
+		if a["matchExpressions"] != b.MatchExpressions {
+			if b.MatchExpressions == nil {
+				return false
+			}
+
+			if len(matchExpressionsA) != len(*b.MatchExpressions) {
+				return false
+			}
+
+			for i := range matchExpressionsA {
+				meA := matchExpressionsA[i]
+				valuesA := meA.(map[string]interface{})["values"].([]interface{})
+				meB := (*b.MatchExpressions)[i]
+
+				if meA.(map[string]interface{})["key"].(string) != meB.Key ||
+					meA.(map[string]interface{})["operator"].(string) != string(meB.Operator) ||
+					!compareStringArrays(valuesA, meB.Values) {
+					return false
+				}
+			}
+		}
+	} else if matchExpressionsA != nil && b.MatchExpressions != nil {
+		return false
+	}
+
+	return true
+}
+
 func TestGetPolicyTemplate(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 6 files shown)


JustinKuli (commented):
This works, and the tests are similar to previous tests, so it has that going for it.

I can think of 2 alternate approaches to the big `compareSelectors` function.
1. "shorter, not better" - use the `map[string]interface{}` to build a string that matches the namespaceSelector's `String()`, and compare those.
2. "too much work" - unmarshal the outputBuffer into an actual type - not `map[string]interface{}`. Locally I started marshalling the first yaml doc back to bytes, then unmarshalled it via sigs.k8s.io/yaml in the actual ConfigurationPolicy type. That seemed like a step in the right direction... but actually it just gets you to two different NamespaceSelector types with the same name, since this repo re-defines the NamespaceSelector type, rather than importing it. I guess you could then use both types' `String()` to get them to the same representation 😆

JustinKuli:
/lgtm
/hold

Feel free to unhold, just giving a chance for other reviews

JustinKuli (approved):

dhaiducek:
Haha, okay I guess I was headed in the right direction with this, then. I thought about redefining the struct so that it's alphabetical and use that to just compare strings as you suggest, but that seemed overly simple for a test. I'm glad this seems like an appropriate approach to you also. It just was starting to seem like I needed a unit test for my helper test function!

internal/utils_test.go
```diff
@@ -30,6 +30,82 @@ func assertReflectEqual(t *testing.T, a interface{}, b interface{}) {
...
+
+	if !compareSelectors(a, b) {
+		t.Fatalf("%s != %s", a, b)
+	}
+}
+
+func compareSelectors(a map[string]interface{}, b types.NamespaceSelector) bool {
+	if includeA, ok := a["include"].([]interface{}); ok {
+		for i := range includeA {
+			if includeA[i].(string) != b.Include[i] {
```
mprahl, line 44: Optional: Since it's just test code, it's fine but this is assuming that `b.Include` has the same length as `includeA`.

internal/utils_test.go
```diff
@@ -30,6 +30,82 @@ func assertReflectEqual(t *testing.T, a interface{}, b interface{}) {
...
+func assertSelectorEqual(t *testing.T, a map[string]interface{}, b types.NamespaceSelector) {
+	t.Helper()
+
+	if !compareSelectors(a, b) {
+		t.Fatalf("%s != %s", a, b)
+	}
+}
+
+func compareSelectors(a map[string]interface{}, b types.NamespaceSelector) bool {
+	if includeA, ok := a["include"].([]interface{}); ok {
```
mprahl, line 42: I think it'd be easier to read if you just converted `a["include"]` to a slice of strings and then perform a reflect deep equal. Then you wouldn't have the concern of the slice lengths not being equal.

The same goes for `exclude`.

internal/utils_test.go
```diff
@@ -30,6 +30,82 @@ func assertReflectEqual(t *testing.T, a interface{}, b interface{}) {
...
+			if excludeA[i].(string) != b.Exclude[i] {
+				return false
+			}
+		}
+	} else if len(b.Exclude) != 0 {
+		return false
+	}
+
+	if matchLabelsA, ok := a["matchLabels"].(map[string]string); ok {
+		if a["matchLabels"] != b.MatchLabels {
```
mprahl, line 63: Would a reflect deep equal work here?

dhaiducek:
Okay, this should be good for another review. Sorry for the noise!

internal/utils_test.go
```diff
@@ -38,38 +38,50 @@ func assertSelectorEqual(t *testing.T, a map[string]interface{}, b types.Namespa
...
+			}
+		}
+
+		if !match {
+			return false
+		}
+	}
+
+	return true
+}
```
mprahl, line 63: It's less efficient but this is less code and easier to understand.

```go
func compareStringArrays(a []interface{}, b []string) bool {
	aTyped := make([]string, len(a))
	for i, val := range a {
		aTyped[i] = val.(string)
	}

	return reflect.DeepEqual(aTyped, b)
}
```

mprahl:
@dhaiducek could you please squash the last three commits together?

internal/utils_test.go
```diff
@@ -30,6 +30,96 @@ func assertReflectEqual(t *testing.T, a interface{}, b interface{}) {
...
+		if !reflect.DeepEqual(matchLabelsA, b.MatchLabels) {
+			return false
+		}
+	} else if matchLabelsA != nil && b.MatchLabels != nil {
+		return false
+	}
+
+	if matchExpressionsA, ok := a["matchExpressions"].([]interface{}); ok {
+		if a["matchExpressions"] != b.MatchExpressions {
+			if b.MatchExpressions != nil {
```
mprahl, line 83: This is very indented, could you please remove one layer by doing this?
```suggestion
			if b.MatchExpressions == nil {
				return false
			}
```

internal/utils_test.go
```diff
@@ -30,6 +30,96 @@ func assertReflectEqual(t *testing.T, a interface{}, b interface{}) {
...
+		if a["matchExpressions"] != b.MatchExpressions {
+			if b.MatchExpressions == nil {
+				return false
+			}
+
+			if len(matchExpressionsA) != len(*b.MatchExpressions) {
+				return false
+			}
+
+			for _, meA := range matchExpressionsA {
```
mprahl, line 91: If I understand this correctly, the reason you are ranging over all of matchExpressionsB for every entry of matchExpressionsA is that you are matching without order mattering?

Do we want to do that?

dhaiducek: Maybe? It is an array of objects where marshalling/unmarshalling could theoretically change the order (though it shouldn't...). You had me thinking of all the ways my comparison could break, so I went HAM on it! I could definitely return it to assuming the order is the same.

mprahl: Marshalling and unmarshalling wouldn't change ordering of an array since order matters, so I would return it to checking the order.

mprahl (approved):

PR #63, opened 2022-08-25
validate resource names for dns compliance

willkutler:

https://github.com/stolostron/backlog/issues/17598

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -13,6 +13,7 @@ import (
 
 	yaml "gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"open-cluster-management.io/ocm-kustomize-generator-plugins/internal/types"
 )
 
@@ -30,6 +31,8 @@ const (
 	placementAPIVersion        = "cluster.open-cluster-management.io/v1beta1"
 	placementKind              = "Placement"
 	maxObjectNameLength        = 63
+	dnsReference               = "https://kubernetes.io/docs/concepts/overview/working-with-objects/names/" +
+		"#dns-subdomain-names"
 )
 
 // Plugin is used to store the PolicyGenerator configuration and the methods to generate the
@@ -648,6 +651,41 @@ func (p *Plugin) assertValidConfig() error {
 		)
 	}
 
+	// validate placement and binding names are DNS compliant
+	defPlrName := p.PolicyDefaults.Placement.PlacementRuleName
+	if defPlrName != "" && len(validation.IsDNS1123Subdomain(defPlrName)) > 0 {
+		return fmt.Errorf(
+			"PolicyDefaults.Placement.PlacementRuleName placement name `%s` is not DNS compliant. See %s",
+			defPlrName,
+			dnsReference,
+		)
+	}
+
+	defPlcmtPlName := p.PolicyDefaults.Placement.PlacementName
+	if defPlcmtPlName != "" && len(validation.IsDNS1123Subdomain(defPlcmtPlName)) > 0 {
+		return fmt.Errorf(
+			"PolicyDefaults.Placement.PlacementName `%s` is not DNS compliant. See %s",
+			defPlcmtPlName,
+			dnsReference,
+		)
+	}
+
+	defPlName := p.PolicyDefaults.Placement.Name
+	if defPlName != "" && len(validation.IsDNS1123Subdomain(defPlName)) > 0 {
+		return fmt.Errorf(
+			"PolicyDefaults.Placement.Name `%s` is not DNS compliant. See %s", defPlName, dnsReference,
+		)
+	}
+
+	if p.PlacementBindingDefaults.Name != "" &&
+		len(validation.IsDNS1123Subdomain(p.PlacementBindingDefaults.Name)) > 0 {
+		return fmt.Errorf(
+			"PlacementBindingDefaults.Name `%s` is not DNS compliant. See %s",
+			p.PlacementBindingDefaults.Name,
+			dnsReference,
+		)
+	}
+
 	defaultPlacementOptions := 0
 	if len(p.PolicyDefaults.Placement.LabelSelector) != 0 || len(p.PolicyDefaults.Placement.ClusterSelectors) != 0 {
 		defaultPlacementOptions++
@@ -682,6 +720,12 @@ func (p *Plugin) assertValidConfig() error {
 		if policy.Name == "" {
 			return fmt.Errorf(
 				"each policy must have a name set, but did not find a name at policy array index %d", i,
+			)
+		}
+
+		if len(validation.IsDNS1123Subdomain(policy.Name)) > 0 {
+			return fmt.Errorf(
+				"policy name `%s` is not DNS compliant. See %s", policy.Name, dnsReference,
 			)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/plugin.go
```diff
@@ -622,6 +637,17 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 	}
 }
 
+// isDNSCompliant verifies that a string is DNS compliant, as is required by some kubernetes resources
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
+func isDNSCompliant(name string) bool {
```
mprahl, line 642: Do you think you could use this function instead since we already depend on this library?
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain

internal/plugin_test.go
```diff
@@ -2467,3 +2467,249 @@ func TestCreatePolicyWithNamespaceSelector(t *testing.T) {
...
+	}
+	basePolicies := p.Policies
+	p.Policies = append(basePolicies, policyConf)
+	p.applyDefaults(map[string]interface{}{})
+
+	if err := p.assertValidConfig(); err != nil {
+		t.Fatal(err.Error())
+	}
+
+	_, err = p.Generate()
```
mprahl, line 2502: Could you make each test in these tests a subtest?

The benefits are:

- It's clearer which test data caused the failure
- The sub tests can run in parallel
- I think it also makes it so that one test failure does not cause the rest of the tests to not run
- We get individual reporting per test case

For example:
https://github.com/stolostron/policy-generator-plugin/blob/49e1893da8f8fc0e1b3c93c3b17b093f3a58aaae/internal/plugin_test.go#L2296-L2316

mprahl:
@willkutler could you also please squash your commits since there is only one "logical" change in the PR?

willkutler:
> @willkutler could you also please squash your commits since there is only one "logical" change in the PR?

@mprahl I was waiting until review to squash so I could add any extra fixes. I added fixes based off your comments so if it looks good to you now I'll squash it.

mprahl (approved):

mprahl (approved):

internal/plugin.go
```diff
@@ -136,6 +139,10 @@ func (p *Plugin) Generate() ([]byte, error) {
 				return nil, err
 			}
 
+			if len(validation.IsDNS1123Subdomain(plcName)) > 0 {
```
mprahl, line 142: In my original review, I didn't realize these checks weren't happening in `assertValidConfig`. Could you please change this?

mprahl (changes requested):

internal/plugin.go
```diff
@@ -648,6 +651,31 @@ func (p *Plugin) assertValidConfig() error {
 		)
 	}
 
+	// validate placement and binding names are DNS compliant
+	placementName := ""
+	if p.PolicyDefaults.Placement.PlacementRuleName != "" {
```
mprahl, line 656: Could you please just validate all these values if they are set so this code doesn't need to be aware of the order of priority?

mprahl, line 656: You'll also need to duplicate this logic for each policy since these values could have been overwritten.

mprahl (changes requested):

willkutler:
@mprahl ready for another review

internal/plugin.go
```diff
@@ -648,6 +651,41 @@ func (p *Plugin) assertValidConfig() error {
 		)
 	}
 
+	// validate placement and binding names are DNS compliant
+	if p.PolicyDefaults.Placement.PlacementRuleName != "" {
+		placementName := p.PolicyDefaults.Placement.PlacementRuleName
+		if placementName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
+			return fmt.Errorf(
+				"default placement name `%s` is not DNS compliant. See %s", placementName, dnsReference,
```
mprahl, line 659:
```suggestion
				"policyDefaults.placement.placementRuleName `%s` is not DNS compliant. See %s", placementName, dnsReference,
```

internal/plugin.go
```diff
@@ -648,6 +651,41 @@ func (p *Plugin) assertValidConfig() error {
...
+			)
+		}
+	}
+	if p.PolicyDefaults.Placement.PlacementName != "" {
+		placementName := p.PolicyDefaults.Placement.PlacementName
+		if placementName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
+			return fmt.Errorf(
+				"default placement name `%s` is not DNS compliant. See %s", placementName, dnsReference,
```
mprahl, line 667:
```suggestion
				"policyDefaults.placement.placementName `%s` is not DNS compliant. See %s", placementName, dnsReference,
```

internal/plugin.go
```diff
@@ -648,6 +651,41 @@ func (p *Plugin) assertValidConfig() error {
...
+			)
+		}
+	}
+	if p.PolicyDefaults.Placement.Name != "" {
+		placementName := p.PolicyDefaults.Placement.Name
+		if placementName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
+			return fmt.Errorf(
+				"default placement name `%s` is not DNS compliant. See %s", placementName, dnsReference,
```
mprahl, line 675:
```suggestion
				"policyDefaults.placement.name `%s` is not DNS compliant. See %s", placementName, dnsReference,
```

internal/plugin.go
```diff
@@ -648,6 +651,41 @@ func (p *Plugin) assertValidConfig() error {
...
+			)
+		}
+	}
+
+	if p.PlacementBindingDefaults.Name != "" &&
+		len(validation.IsDNS1123Subdomain(p.PlacementBindingDefaults.Name)) > 0 {
+		return fmt.Errorf(
+			"default placement binding name `%s` is not DNS compliant. See %s",
```
mprahl, line 683:
```suggestion
			"placementBindingDefaults.name `%s` is not DNS compliant. See %s",
```

internal/plugin.go
```diff
@@ -799,6 +843,32 @@ func (p *Plugin) assertValidConfig() error {
 			)
 		}
 
+		// validate placement names are DNS compliant
+		if policy.Placement.PlacementRuleName != "" {
+			placementName := p.PolicyDefaults.Placement.PlacementRuleName
```
mprahl, line 848: This should be `policy.Placement.PlacementRuleName` right?

internal/plugin.go
```diff
@@ -799,6 +843,32 @@ func (p *Plugin) assertValidConfig() error {
...
+			if placementName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
+				return fmt.Errorf(
+					"default placement name `%s` is not DNS compliant. See %s", placementName, dnsReference,
+				)
+			}
+		}
+		if policy.Placement.PlacementName != "" {
+			placementName := p.PolicyDefaults.Placement.PlacementName
```
mprahl, line 856: This should be `policy.Placement.PlacementName` right?

internal/plugin.go
```diff
@@ -799,6 +843,32 @@ func (p *Plugin) assertValidConfig() error {
...
+			if placementName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
+				return fmt.Errorf(
+					"default placement name `%s` is not DNS compliant. See %s", placementName, dnsReference,
+				)
+			}
+		}
+		if policy.Placement.Name != "" {
+			placementName := p.PolicyDefaults.Placement.Name
```
mprahl, line 864: This should be `policy.Placement.Name` right?

internal/plugin.go
```diff
@@ -648,6 +651,41 @@ func (p *Plugin) assertValidConfig() error {
 		)
 	}
 
+	// validate placement and binding names are DNS compliant
+	if p.PolicyDefaults.Placement.PlacementRuleName != "" {
+		placementName := p.PolicyDefaults.Placement.PlacementRuleName
+		if placementName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
```
mprahl, line 657: Could you combine this with the previous if statements? Same goes for the others below.

mprahl (changes requested):

internal/plugin.go
willkutler: what do you mean here? I split them into separate if statements since we didn't want to prioritize certain fields

mprahl: Oh, I meant:
```go
if p.PolicyDefaults.Placement.PlacementRuleName != "" && len(validation.IsDNS1123Subdomain(placementName)) > 0 {
```

mprahl (approved):
Awesome!

PR #64, opened 2022-08-29
Fail generation and add help on field typos

JustinKuli:
When the input configuration does not match the schema in the go type,
the generator will now respond with an error. The intention is to help
users when there is a typo in the configuration, which otherwise would
just be ignored (and in the case of a map, any configuration "inside"
would be dropped).

This also adds help text to these error cases, to try and identify what
field the user might have meant. That implementation uses a library
which does not compute the Levenshtein distance, but should work well
enough, and has the advantage of already being in the import graph.

Refs:
 - https://github.com/stolostron/backlog/issues/19646

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -38,7 +38,9 @@ const (
 // Plugin is used to store the PolicyGenerator configuration and the methods to generate the
 // desired policies.
 type Plugin struct {
-	Metadata struct {
+	APIVersion string `json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"`
+	Kind       string `json:"kind,omitempty" yaml:"kind,omitempty"`
+	Metadata   struct {
 		Name string `json:"name,omitempty" yaml:"name,omitempty"`
 	} `json:"metadata,omitempty" yaml:"metadata,omitempty"`
 	PlacementBindingDefaults struct {
@@ -75,11 +77,14 @@ var defaults = types.PolicyDefaults{
 // Config validates the input PolicyGenerator configuration, applies any missing defaults, and
 // configures the Policy object.
 func (p *Plugin) Config(config []byte, baseDirectory string) error {
-	err := yaml.Unmarshal(config, p)
+	dec := yaml.NewDecoder(bytes.NewReader(config))
+	dec.KnownFields(true) // emit an error on unknown fields in the input
+
+	err := dec.Decode(p)
 	const errTemplate = "the PolicyGenerator configuration file is invalid: %w"
 
 	if err != nil {
-		return fmt.Errorf(errTemplate, err)
+		return fmt.Errorf(errTemplate, addFieldNotFoundHelp(err))
 	}
 
 	var unmarshaledConfig map[string]interface{}
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module open-cluster-management.io/ocm-kustomize-generator-plugins
 go 1.18
 
 require (
+	github.com/pmezard/go-difflib v1.0.0
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/apimachinery v0.23.3
@@ -32,7 +33,6 @@ require (
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/monochromegane/go-gitignore v0.0.0-20200626010858-205db1a8cc00 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/stretchr/testify v1.7.0 // indirect
 	github.com/xlab/treeprint v1.1.0 // indirect
 	go.starlark.net v0.0.0-20220213143740-c55a923347b1 // indirect
--- a/internal/typohelper.go
+++ b/internal/typohelper.go
@@ -1,0 +1,118 @@
+package internal
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"regexp"
+	"strings"
+
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// addFieldNotFoundHelp adds help text to errors that occur when the input config
+// has a field that is not present in the Config struct (eg a typo), to assist
+// the user in debugging the problem. If the input error was not from a missing
+// field, then it is returned unchanged.
+func addFieldNotFoundHelp(err error) error {
+	re := regexp.MustCompile(`field (\S*) not found in type (\S*)`)
+
+	repl := func(line string) string {
+		match := re.FindStringSubmatch(line)
+
+		fieldType := reflect.TypeOf(Plugin{})
+		fieldTag := "PolicyGenerator"
+
+		if match[2] != fieldType.String() {
+			// Search the right type, if it's not the top-level object.
+			fieldType, fieldTag = findNestedType(fieldType, match[2], "")
+		}
+
+		msg := fmt.Sprintf("field %v found but not defined", match[1])
+
+		if fieldTag != "" {
+			msg += fmt.Sprintf(" in type %v", fieldTag)
+		}
+
+		suggestion := autocorrectField(match[1], fieldType)
+		if suggestion != "" {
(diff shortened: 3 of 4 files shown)


JustinKuli:
Just a draft for now, I want to add a test, but I want to make sure the direction I'm taking is not the wrong one.

With this change, running the generator directly with this config in the examples directory:
```yaml
apiVersion: policy.open-cluster-management.io/v1
kind: PolicyGenerator
metadata:
  name: policy-generator-name
placementBindingDefaults:
  name: my-placement-binding
policyDefaults:
  controls: 
    - PR.DS-1 Data-at-rest
  namespace: my-policies
  placement:
    placementRulePath: input/placementrule.yaml
  remediationAction: inform
  severity: medium
  policySet: 
    - policyset-kyverno
policies:
- name: policy-app-config-aliens
  disabled: false
  manifests:
    - path: input/configmap-aliens.yaml
      patches:
        - apiVersion: v1
          kind: ConfigMap
          metadata:
            labels:
              chandler: bing
  remediationAction: enforce
policySets:
- name: policyset-kyverno
  description: this is a kyverno policy set.
  policies: []
  placement:
    placementRulePath: input/placementrule.yaml
```

Produces this error:
```
error processing the PolicyGenerator file 'policyGenerator.yaml': the PolicyGenerator configuration file is invalid: yaml: unmarshal errors:
  line 15: field policySet not found in type types.PolicyDefaults - did you mean 'policySets'?
```

mprahl:
> Just a draft for now, I want to add a test, but I want to make sure the direction I'm taking is not the wrong one.
> 
> With this change, running the generator directly with this config in the examples directory:
> 
> ```yaml
> apiVersion: policy.open-cluster-management.io/v1
> kind: PolicyGenerator
> metadata:
>   name: policy-generator-name
> placementBindingDefaults:
>   name: my-placement-binding
> policyDefaults:
>   controls: 
>     - PR.DS-1 Data-at-rest
>   namespace: my-policies
>   placement:
>     placementRulePath: input/placementrule.yaml
>   remediationAction: inform
>   severity: medium
>   policySet: 
>     - policyset-kyverno
> policies:
> - name: policy-app-config-aliens
>   disabled: false
>   manifests:
>     - path: input/configmap-aliens.yaml
>       patches:
>         - apiVersion: v1
>           kind: ConfigMap
>           metadata:
>             labels:
>               chandler: bing
>   remediationAction: enforce
> policySets:
> - name: policyset-kyverno
>   description: this is a kyverno policy set.
>   policies: []
>   placement:
>     placementRulePath: input/placementrule.yaml
> ```
> 
> Produces this error:
> 
> ```
> error processing the PolicyGenerator file 'policyGenerator.yaml': the PolicyGenerator configuration file is invalid: yaml: unmarshal errors:
>   line 15: field policySet not found in type types.PolicyDefaults - did you mean 'policySets'?
> ```

This is cool. The issue is that `types.PolicyDefaults` means nothing to the user since that is a Go construct. You could add logic so that it converts `types.PolicyDefaults` to `policyDefaults`.

mprahl:
@JustinKuli I think we could simplify this to just point to the path in the YAML that is invalid rather than supply a suggestion. What do you think?

JustinKuli:
Partly I added the suggestion to make the change more than just 3 lines 😉 ... but I also do think it is pretty helpful. Especially in this example case, where the user is just missing the `s`, without a suggestion I might look back and forth between my YAML and the example multiple times and never see the difference.

> You could add logic so that it converts `types.PolicyDefaults` to `policyDefaults`.

That could be good. While we're at it, I would consider rephrasing more of the error, so it's more like:
```
line 15: field policySet found but not defined in type policyDefaults - did you mean 'policySets'?
```

mprahl:
> Partly I added the suggestion to make the change more than just 3 lines wink ... but I also do think it is pretty helpful. Especially in this example case, where the user is just missing the `s`, without a suggestion I might look back and forth between my YAML and the example multiple times and never see the difference.
> 
> > You could add logic so that it converts `types.PolicyDefaults` to `policyDefaults`.
> 
> That could be good. While we're at it, I would consider rephrasing more of the error, so it's more like:
> 
> ```
> line 15: field policySet found but not defined in type policyDefaults - did you mean 'policySets'?
> ```
> 
> What do you think?

That sounds good.

dhaiducek (commented):
This is neat! The suggestion is pretty nice so that users may not have to go digging for what it should have been. This PR also means we need to be much more disciplined about adding new fields to the generator (which is good)!

internal/plugin_test.go
```diff
@@ -2663,3 +2665,164 @@ func TestGenerateNonDNSBindingName(t *testing.T) {
...
+  namespace: minimal
+policies:
+- name: my-minimal
+  manifests:
+  - paths: @@@
+`),
+		},
+		"typo no suggestion": {
+			desiredErrs: []string{
+				"line 12: field namespace found but not defined in type manifests$",
```
dhaiducek, line 2768: Not an issue and more of a curiosity question: What's the meaning of the trailing `$`?

dhaiducek (approved):
/hold for other reviews

Some fancy processing, but it's probably worth surfacing the references and helpful errors since generic Kustomize gives very little direction there.

internal/plugin.go
```diff
@@ -101,6 +109,113 @@ func (p *Plugin) Config(config []byte, baseDirectory string) error {
 	return p.assertValidConfig()
 }
 
+// addFieldNotFoundHelp adds help text to errors that occur when the input config
+// has a field that is not present in the Config struct (eg a typo), to assist
+// the user in debugging the problem. If the input error was not from a missing
+// field, then it is returned unchanged.
+func addFieldNotFoundHelp(err error) error {
```
mprahl, line 116: Could you please add these functions to a new file so that `plugin.go` doesn't keep growing?

internal/plugin.go
```diff
@@ -101,6 +109,113 @@ func (p *Plugin) Config(config []byte, baseDirectory string) error {
...
+	if helpMsg == err.Error() {
+		// Error was unchanged, return the original to preserve the type
+		return err
+	}
+
+	return errors.New(helpMsg)
+}
+
+// autocorrectField returns the field in the containingType which most closely
+// matches the input badType. It will return an empty string if no match can
```
mprahl, line 155:
```suggestion
// matches the input badField. It will return an empty string if no match can
```

internal/plugin.go
```diff
@@ -101,6 +109,113 @@ func (p *Plugin) Config(config []byte, baseDirectory string) error {
...
+		return ""
+	}
+
+	matcher := difflib.NewMatcher(strings.Split(badField, ""), []string{""})
+	bestRatio := 0.85 // require 85% or better match
+	bestMatch := ""
+
+	// iterate over all fields in the struct that have a yaml tag.
+	for _, field := range reflect.VisibleFields(containingType) {
+		yamlTag := strings.TrimSuffix(field.Tag.Get("yaml"), ",omitempty")
```
mprahl, line 168: Optional: I think it'd be more resilient if you did something like this:
```suggestion
		yamlTag := strings.TrimSuffix(strings.SplitN(field.Tag.Get("yaml"), ",", 2)[0])
```

internal/plugin.go
```diff
@@ -101,6 +109,113 @@ func (p *Plugin) Config(config []byte, baseDirectory string) error {
...
+		return findNestedType(baseType.Elem(), want, tag)
+	}
+
+	if baseType.Kind() != reflect.Struct {
+		return nil, ""
+	}
+
+	// iterate over all fields in the struct that have a yaml tag.
+	for _, field := range reflect.VisibleFields(baseType) {
+		yamlTag := strings.TrimSuffix(field.Tag.Get("yaml"), ",omitempty")
```
mprahl, line 204: Optional: Same comment as above about using `SplitN`.

mprahl (changes requested):
Thanks for implementing this! This will save me some time from helping Christian debug typos in his Policy Generator config files.  ;)

internal/plugin.go
JustinKuli: Good point

mprahl (approved):

PR #65, opened 2022-08-31
Support Kustomize directories

dhaiducek:
Addresses:
- https://github.com/stolostron/backlog/issues/25290

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 
 	"github.com/spf13/pflag"
 	"open-cluster-management.io/ocm-kustomize-generator-plugins/internal"
@@ -67,7 +66,7 @@ func processGeneratorConfig(filePath string) []byte {
 		errorAndExit("failed to read file '%s': %s", filePath, err)
 	}
 
-	err = p.Config(fileData, path.Dir(cwd))
+	err = p.Config(fileData, cwd)
 	if err != nil {
 		errorAndExit("error processing the PolicyGenerator file '%s': %s", filePath, err)
 	}
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -122,8 +122,9 @@ policies:
     manifests:
         # Required. Path to a single file or a flat directory of files relative to the
         # kustomization.yaml file. This path cannot be in a directory outside of the directory with
-        # the kustomization.yaml file. Subdirectories within the directory with kustomization.yaml
-        # file are allowed.
+        # the kustomization.yaml file. Subdirectories within the directory of the kustomization.yaml
+        # file are allowed. Kustomization subdirectories are also supported and will not process any
+        # YAML files in the subdirectory if a kustomization.yaml file is found.
         # Supported manifests:
         #   1) Non-root policy type manifests such as IamPolicy, CertificatePolicy, and ConfigurationPolicy
         #      that have a "Policy" suffix. These are not modified except for patches and are directly added 
@@ -137,10 +138,10 @@ policies:
         metadataComplianceType: ""
          # Optional. (See policyDefaults.evaluationInterval for description.)
         evaluationInterval: {}
-        # Optional. A Kustomize patch to apply to the manifest(s) at the path. If there
-        # are multiple manifests, the patch requires the apiVersion, kind, metadata.name,
-        # and metadata.namespace (if applicable) fields to be set so Kustomize
-        # can identify the manifest the patch applies to.
+        # (Note: a path to a directory containing a Kustomize manifest is a supported alternative.) Optional. A
+        # Kustomize patch to apply to the manifest(s) at the path. If there are multiple manifests, the patch requires
+        # the apiVersion, kind, metadata.name, and metadata.namespace (if applicable) fields to be set so Kustomize can
+        # identify the manifest the patch applies to.
         patches:
             # Optional: Only required when there are multiple manifests in the path.
           - apiVersion: ""
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -16,6 +16,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"open-cluster-management.io/ocm-kustomize-generator-plugins/internal/expanders"
 	"open-cluster-management.io/ocm-kustomize-generator-plugins/internal/types"
+	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
 // getManifests will get all of the manifest files associated with the input policy configuration
@@ -23,6 +25,7 @@ import (
 // be read.
 func getManifests(policyConf *types.PolicyConfig) ([][]map[string]interface{}, error) {
 	manifests := [][]map[string]interface{}{}
+	hasKustomize := map[string]bool{}
 
 	for _, manifest := range policyConf.Manifests {
 		manifestPaths := []string{}
@@ -33,6 +36,8 @@ func getManifests(policyConf *types.PolicyConfig) ([][]map[string]interface{}, e
 		if err != nil {
 			return nil, readErr
 		}
+
+		resolvedFiles := []string{}
 
 		if manifestPathInfo.IsDir() {
 			files, err := ioutil.ReadDir(manifest.Path)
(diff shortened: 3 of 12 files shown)


dhaiducek:
Investigation is being done to implement the generator as a transformer rather than wrapping Kustomize (again) 😄

cmd/main.go
```diff
@@ -67,7 +66,7 @@ func processGeneratorConfig(filePath string) []byte {
 		errorAndExit("failed to read file '%s': %s", filePath, err)
 	}
 
-	err = p.Config(fileData, path.Dir(cwd))
+	err = p.Config(fileData, cwd)
```
dhaiducek, line 69: I'm not sure why we stripped out the last bit of the path (which is the actual working directory). But it was preventing identifying when a `kustomization.yaml` pointed back to itself: https://github.com/stolostron/policy-generator-plugin/pull/65/commits/c0158ef76fbe2c1c94a503e2506869b9eb9b849d

@mprahl Do you recall why we did it this way? The tests pass without it, which is also confusing.

internal/utils_test.go
```diff
@@ -1152,3 +1166,66 @@ func TestVerifyManifestPath(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create the directory structure %s: %v", kustomizeDir, err)
+	}
+
+	// Create files in baseDirectory/kustomizedir
+	manifestPaths := map[string]string{
+		"kustomization.yaml": `
+resources:
+- configmap.yaml
+- https://github.com/dhaiducek/policy-generator-plugin/examples/input-kustomize/?ref=support-kustomize
```
dhaiducek, line 1191: This is where the peer example PR comes in:
- #66

docs/policygenerator-reference.yaml
```diff
@@ -137,10 +138,10 @@ policies:
         metadataComplianceType: ""
          # Optional. (See policyDefaults.evaluationInterval for description.)
         evaluationInterval: {}
-        # Optional. A Kustomize patch to apply to the manifest(s) at the path. If there
-        # are multiple manifests, the patch requires the apiVersion, kind, metadata.name,
-        # and metadata.namespace (if applicable) fields to be set so Kustomize
-        # can identify the manifest the patch applies to.
+        # (deprecated) Optional. A Kustomize patch to apply to the manifest(s) at the path. If
```
dhaiducek, line 141: I also figured in-config patches would be deprecated with this addition.

JustinKuli: I definitely think using another kustomization.yaml to apply the patches is a better solution, so I think we should encourage that somehow. To me, saying it's "deprecated" implies we plan on removing it - so do we? What would the timeline be?

If we don't plan on actually removing the option, I think we could just suggest that using another kustomization.yaml file is a better solution.

internal/utils_test.go
JustinKuli: Ah, after 66 is merged, this would be updated to point to the stolostron repo (and main branch)?

JustinKuli (approved):
Overall, it looks good to me!

internal/utils_test.go
dhaiducek: Correct. I thought about setting up another repo for it, but that seemed unnecessary.

JustinKuli:
One concern that had come up from us calling kustomize from our plugin (which is itself called from kustomize) was "what if one of the kustomizations we try to use ends up calling another plugin, and we get into a loop that kustomize can't recognize?" (Note that @dhaiducek reported that kustomize will notice if kustomization.yaml files end up calling each other in a loop). We shouldn't need to worry about that - no plugins can be called without enabling alpha plugins explicitly. See https://kubectl.docs.kubernetes.io/guides/extending_kustomize/#required-alpha-flags

Based on the above doc, the only plugins that would be able to run in the ACM application environment is ones that are specially installed (like this plugin). No "in-repo" exec plugins would be able to run without the `--enable-exec` flag, and container plugins would likely not be able to run without a container runtime.

docs/policygenerator-reference.yaml
dhaiducek: I've added a note to the deprecation to use a Kustomize directory instead.

cmd/main.go
mprahl: I don't remember but I don't see a reason why it's like that. So I'm +1 to your change.

internal/utils_test.go
```diff
@@ -253,6 +253,156 @@ data:
 	}
 }
 
+func TestGetPolicyTemplateKustomize(t *testing.T) {
+	t.Parallel()
+	kustomizeDir := t.TempDir()
+	configStrings := []string{"tomato", "potato"}
+
+	manifestsDir := path.Join(kustomizeDir, "manifests")
```
mprahl, line 261: What is `manifestsDir` used for in this test?

internal/utils_test.go
```diff
@@ -1152,3 +1316,66 @@ func TestVerifyManifestPath(t *testing.T) {
...
+	for filename, content := range manifestPaths {
+		manifestPath := path.Join(kustomizeDir, filename)
+
+		err = ioutil.WriteFile(manifestPath, []byte(content), 0o666)
+		if err != nil {
+			t.Fatalf("Failed to write %s", manifestPath)
+		}
+	}
+
+	t.Run(
```
mprahl, line 1365: Does this need to run in a `t.Run` since there aren't multiple testcases?

mprahl (changes requested):

internal/utils_test.go
dhaiducek: It's the directory called in the Kustomize file.

internal/utils_test.go
dhaiducek: Indeed it does not 🙂

internal/utils.go
```diff
@@ -245,6 +270,28 @@ func setNamespaceSelector(policyConf *types.PolicyConfig, policyTemplate *map[st
 	}
 }
 
+// processKustomizeDir runs a provided directory through Kustomize in order to generate the manifests within it.
+func processKustomizeDir(path string) (*[]map[string]interface{}, error) {
+	k := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+
+	resourceMap, err := k.Run(filesys.MakeFsOnDisk(), path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to process provided kustomize directory: %w", err)
```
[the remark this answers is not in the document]
dhaiducek: It does, but it would be good to put it up front, especially since it can give quite a trace:
```
error generating policies from the PolicyGenerator file '/var/folders/mt/6766yzb50jlb9t0x6gvl0nz00000gn/T/kust-plugin-config-184116644': failed to process provided kustomize directory: accumulating resources: accumulation err='accumulating resources from 'dev': '/Users/dhaiduce/github/open-cluster-management/grc/extras/policy-generator-plugin/examples/input-kustomize/dev' must resolve to a file': recursed accumulation of path '/Users/dhaiduce/github/open-cluster-management/grc/extras/policy-generator-plugin/examples/input-kustomize/dev': accumulating resources: accumulation err='accumulating resources from '../base': '/Users/dhaiduce/github/open-cluster-management/grc/extras/policy-generator-plugin/examples/input-kustomize/base' must resolve to a file': recursed accumulation of path '/Users/dhaiduce/github/open-cluster-management/grc/extras/policy-generator-plugin/examples/input-kustomize/base': accumulating resources: accumulation err='accumulating resources from 'nginx-pod.yaml': yaml: line 8: mapping values are not allowed in this context': got file 'nginx-pod.yaml', but '/Users/dhaiduce/github/open-cluster-management/grc/extras/policy-generator-plugin/examples/input-kustomize/base/nginx-pod.yaml' must be a directory to be a root
```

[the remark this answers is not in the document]
dhaiducek: Side note: It'd be nice if we could figure out how to reference the original file instead of the tmp generator file. That might not be possible, though.

docs/policygenerator-reference.yaml
```diff
@@ -137,10 +138,10 @@ policies:
         metadataComplianceType: ""
          # Optional. (See policyDefaults.evaluationInterval for description.)
         evaluationInterval: {}
-        # Optional. A Kustomize patch to apply to the manifest(s) at the path. If there
-        # are multiple manifests, the patch requires the apiVersion, kind, metadata.name,
-        # and metadata.namespace (if applicable) fields to be set so Kustomize
-        # can identify the manifest the patch applies to.
+        # (deprecated - Use a path to a directory containing a Kustomize manifest instead) Optional. A Kustomize patch
```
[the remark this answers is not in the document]
dhaiducek: Ah, I didn't realize that. I've updated the message for now.

mprahl (approved):

PR #69, opened 2022-09-14
Support `PruneObjectBehavior`

dhaiducek:
I was uncertain whether the value should be validated. I didn't validate it since other string fields in the generator (like `complianceType` and `remediationAction`) didn't appear to have validation in place.

Addresses:
- https://github.com/stolostron/backlog/issues/25353

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -43,6 +43,10 @@ policyDefaults:
     # avoid evaluating the policy after it has become a particular compliance state.
     compliant: 30m
     noncompliant: 45s
+  # Optional. Determines whether objects created or monitored by the policy should be deleted when the policy is
+  # deleted. Pruning only takes place if the remediation action of the policy has been set to "enforce". Example values
+  # are "DeleteIfCreated", "DeleteAll", or "None". This defaults to unset, which is equivalent to "None".
+  pruneObjectBehavior: "None"
   # Optional. When the policy references a Kyverno policy manifest, this determines if an additional
   # configuration policy should be generated in order to receive policy violations in Open Cluster
   # Management when the Kyverno policy has been violated. This defaults to true.
@@ -138,6 +142,8 @@ policies:
         metadataComplianceType: ""
          # Optional. (See policyDefaults.evaluationInterval for description.)
         evaluationInterval: {}
+        # Optional. (See policyDefaults.pruneObjectBehavior for description.)
+        pruneObjectBehavior: ""
         # (Note: a path to a directory containing a Kustomize manifest is a supported alternative.) Optional. A
         # Kustomize patch to apply to the manifest(s) at the path. If there are multiple manifests, the patch requires
         # the apiVersion, kind, metadata.name, and metadata.namespace (if applicable) fields to be set so Kustomize can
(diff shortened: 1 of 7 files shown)


docs/policygenerator-reference.yaml
```diff
@@ -43,6 +43,10 @@ policyDefaults:
     # avoid evaluating the policy after it has become a particular compliance state.
     compliant: 30m
     noncompliant: 45s
+  # Optional. Determines whether objects created or monitored by the policy should be deleted when the policy is
```
ChunxiAlexLuo, line 46: Not sure if the difference between "DeleteIfCreated", "DeleteAll" so be mentioned here, the rest looks good to me

ChunxiAlexLuo (approved):
/lgtm

docs/policygenerator-reference.yaml
```diff
@@ -43,6 +43,10 @@ policyDefaults:
...
+  # deleted. Pruning only takes place if the remediation action of the policy has been set to "enforce". Accepts values
```
mprahl, line 47: So we don't have to update this doc every time a new option is supported in the API.
```suggestion
  # deleted. Pruning only takes place if the remediation action of the policy has been set to "enforce". Example values
```

internal/plugin.go
```diff
@@ -463,6 +463,10 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			}
 		}
 
+		if policy.PruneObjectBehavior == "" {
```
mprahl, line 466: You'll also need to handle this at the manifest level (i.e. under `for j := range policy.Manifests {`).

docs/policygenerator-reference.yaml
dhaiducek: It's a good thought, but I think I'd prefer to keep these descriptions on the shorter side, as with `complianceType`. And then, as Matt pointed out below, if another value comes along we don't have to feel like we need to add it here.

internal/plugin.go
dhaiducek: So, this is the point I was trying to make in our scrum. If we're passing `pruneObjectBehavior` to `manifests`, we should be passing everything (`remediationAction`, `severity`, `namespaceSelector`, and so on), right?

dhaiducek: (I've added it into `manifests` for now.)

mprahl: Right. I'm on the fence but I'm leaning towards adding support for those at the manifest level (i.e. ConfigurationPolicy) since we already do for some fields, so the ship has sailed on that so being consistent would be good. If we do go down this path, could you please open a backlog issue for this?

It'd be good to clarify that this would only apply if `consolidateManifests` is `false`.

What do you think?

internal/utils.go
```diff
@@ -281,6 +281,11 @@ func buildPolicyTemplate(
 		metadata["annotations"] = policyConf.ConfigurationPolicyAnnotations
 	}
 
+	if policyConf.PruneObjectBehavior != "" {
```
mprahl, line 284: This would need to come from the manifest instead.

internal/plugin.go
dhaiducek: Works for me! I've created the issue: https://github.com/stolostron/backlog/issues/25982

dhaiducek:
Okay, this PR should be ready for another review.

mprahl (approved):

PR #71, opened 2022-10-04
Consolidate structs / Consistent `manifests` overrides

dhaiducek:
This PR is twofold:
- Implement `PolicyOptions` and `ConfigurationPolicyOptions`, which folds keys into a single struct that can be used in the `PolicyDefaults`, `Policy`, and `Manifests` structs.
- The consolidation brings consistency to overrides in `Manifests` that was missing previously, so this also adds the required validation and overrides there


mprahl (approved):
Really nice job on this!

PR #73, opened 2022-10-06
Remove unnecessary pointers

dhaiducek:


mprahl (approved):
Much cleaner!

I personally would have kept the structs as pointers to avoid copying all the data fields when passing them in a function/method but it's not a big deal.

/hold in case you want to change that or others want to review. Feel free to unhold.

mprahl (approved):

PR #80, opened 2022-10-20
Add release workflow

dhaiducek:
- https://github.com/stolostron/backlog/issues/26085

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,35 @@
+# Copyright Contributors to the Open Cluster Management project
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+name: Create Release
+
+jobs:
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+
+      - name: Build project
+        run: |
+          make build-release
+
+      - name: Release
+        uses: softprops/action-gh-release@v0.1.14
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          files: build_output/*-PolicyGenerator*
+          generate_release_notes: true
+          draft: true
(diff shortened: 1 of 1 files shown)


dhaiducek:
This is largely taken from `clusteradm`'s workflow and will generate a draft release whenever a tag is pushed. I've updated the Policy Generator release instructions accordingly: https://docs.google.com/document/d/1puzltjSbIoNiNA9uMhWLOZGyYPj69ZGIhvnxTo0ihc4/edit#

.github/release.yml
```diff
@@ -0,0 +1,36 @@
...
+
+name: Create Release
+
+jobs:
+  release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
mprahl, line 16: Optional:
```suggestion
        uses: actions/checkout@v3
```

.github/release.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
```
mprahl, line 19:
```suggestion
        uses: actions/setup-go@v3
        with:
          go-version-file: go.mod
```

.github/release.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Build project
+        run: |
+          make build-binary
```
mprahl, line 23:
```suggestion
          make build-release
```

.github/release.yml
dhaiducek: There is no `v3` for `checkout`: https://github.com/actions/checkout/releases

.github/release.yml
dhaiducek: Ah, missed this one--thanks!

dhaiducek:
Sorry--the file was in the `.github` folder instead of `.github/workflows` and I moved it, updated the wrong file, and then deleted the updated file. 🤦 Should be good now. 🙂

.github/release.yml
mprahl: @dhaiducek there is, but v2 is also supported still:
https://github.com/actions/checkout/releases/tag/v3.1.0

mprahl (approved):

PR #81, opened 2022-10-27
OCPBUGS-2200: Add remediationAction at the Policy spec level

tliu2021:
Signed-off-by: Tao Liu <tali@redhat.com>

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -486,3 +486,22 @@ func verifyManifestPath(baseDirectory string, manifestPath string) error {
 
 	return nil
 }
+
+// Check policy-templates to see if all the remediation actions match, if so return the root policy remediation action
+func getRootRemediationAction(policyTemplates []map[string]map[string]interface{}) string {
+	var action string
+
+	for _, value := range policyTemplates {
+		if spec, ok := value["objectDefinition"]["spec"].(map[string]interface{}); ok {
+			if _, ok = spec["remediationAction"].(string); ok {
+				if action == "" {
+					action = spec["remediationAction"].(string)
+				} else if spec["remediationAction"].(string) != action {
+					return ""
+				}
+			}
+		}
+	}
+
+	return action
+}
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -1167,6 +1167,11 @@ func (p *Plugin) createPolicy(policyConf *types.PolicyConfig) error {
 		},
 	}
 
+	// set the root policy remediation action if all the remediation actions match
+	if rootRemediationAction := getRootRemediationAction(policyTemplates); rootRemediationAction != "" {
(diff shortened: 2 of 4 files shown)


internal/utils.go
```diff
@@ -486,3 +486,20 @@ func verifyManifestPath(baseDirectory string, manifestPath string) error {
...
 	return nil
 }
+
+// Check policy-templates to see if all the remediation actions match, if so set the root policy remediation action
+func getRootRemediationAction(policyTemplates []map[string]map[string]interface{}) string {
+
+	var action string
+	for i, value := range policyTemplates {
+		spec := value["objectDefinition"]["spec"].(map[string]interface{})
+		if i == 0 {
```
mprahl, line 496: Optional: You could avoid using `i` if you changed this to `if action == ""`.

mprahl:
@tliu2021 thanks so much for the contribution! Could you please address the linting failures?

Also, could you please restructure your commit message to follow our standards?
https://github.com/open-cluster-management-io/community/blob/main/sig-policy/contribution-guidelines.md#commit-messages

You can look at the following commits as examples:
https://github.com/stolostron/policy-generator-plugin/commit/88a386a35a2961edeb1a8cb9120e412745166173
https://github.com/stolostron/policy-generator-plugin/commit/cde84ee52ff69df01d1fbbdc37acaaf5332b5c5a

tliu2021:
> 

The commit message has been updated and linting failures are fixed.

internal/utils.go
```diff
@@ -486,3 +486,19 @@ func verifyManifestPath(baseDirectory string, manifestPath string) error {
 
 	return nil
 }
+
+// Check policy-templates to see if all the remediation actions match, if so set the root policy remediation action
```
mprahl, line 490: Using the word "set" implies a change is made to the input data, but that's not the case.
```suggestion
// Check policy-templates to see if all the remediation actions match, if so return the root policy remediation action
```

internal/utils.go
```diff
@@ -486,3 +486,19 @@ func verifyManifestPath(baseDirectory string, manifestPath string) error {
...
+func getRootRemediationAction(policyTemplates []map[string]map[string]interface{}) string {
+	var action string
+
+	for _, value := range policyTemplates {
+		spec := value["objectDefinition"]["spec"].(map[string]interface{})
```
mprahl, line 495: In the case where an `IAMPolicy` or `CertificatePolicy` is specified as a manifest, then `policy-template` entry is not generated and thus not guaranteed to have `spec` and `spec.remediationAction` set, so you need to account for bad user input here. You can do this by verifying the type assertions.

internal/utils.go
tliu2021: changed.

mprahl (approved):
This looks great. Thanks for the contribution!

PR #82, opened 2022-11-01
Add options for policy ordering

JustinKuli:
The new `dependencies` field in the generator is the same as the new `dependencies` field in the Policy type, however, more fields are optional and will be filled in by the generator with useful defaults.

The new `ignorePending` field is also the same, but since that would be set on templates, and the generator does not expose much at the template level, here the field is set on the policy, and the generator sets the correct field in all the templates of that policy. For a similar reason, the `extraDependencies` field is not exposed in the generator at all.

The new `orderViaDependencies` field will set up a chain of dependencies on the created policies. So if 3 policies are created by the generator, policy 2 will depend on policy 1, and policy 3 will depend on policy 2.

In all cases, dependency lists are merged - setting it on a specific policy will *not* override the defaults, or a dependency added by the `orderViaDependencies` flag.

Refs:
 - https://github.com/stolostron/backlog/issues/26183

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -46,6 +46,8 @@ type Plugin struct {
 	PlacementBindingDefaults struct {
 		Name string `json:"name,omitempty" yaml:"name,omitempty"`
 	} `json:"placementBindingDefaults,omitempty" yaml:"placementBindingDefaults,omitempty"`
+	OrderViaDependencies bool `json:"orderViaDependencies,omitempty" yaml:"orderViaDependencies,omitempty"`
+
 	PolicyDefaults types.PolicyDefaults    `json:"policyDefaults,omitempty" yaml:"policyDefaults,omitempty"`
 	Policies       []types.PolicyConfig    `json:"policies" yaml:"policies"`
 	PolicySets     []types.PolicySetConfig `json:"policySets" yaml:"policySets"`
@@ -411,6 +413,27 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 		}
 	}
 
+	for i, dep := range p.PolicyDefaults.Dependencies {
+		if dep.Namespace == "" {
+			p.PolicyDefaults.Dependencies[i].Namespace = p.PolicyDefaults.Namespace
+		}
+
+		if dep.Compliance == "" {
+			p.PolicyDefaults.Dependencies[i].Compliance = "Compliant"
+		}
+
+		if dep.Kind == "" {
+			p.PolicyDefaults.Dependencies[i].Kind = policyKind
+		}
+
+		if dep.APIVersion == "" {
+			p.PolicyDefaults.Dependencies[i].APIVersion = policyAPIVersion
+		}
+	}
+
+	// Used when p.OrderViaDependencies is set
+	prevPolicyName := ""
+
 	for i := range p.Policies {
 		policy := &p.Policies[i]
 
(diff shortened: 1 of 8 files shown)


internal/plugin_test.go
```diff
@@ -4,7 +4,7 @@ package internal
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
```
willkutler, line 7: why the change in imports?

willkutler (approved):

willkutler:
looks good, just curious about the import change

internal/plugin_test.go
dhaiducek: `ioutil` is deprecated: https://pkg.go.dev/io/ioutil

willkutler: gotcha, thanks!

willkutler:
oops, I didn't realize there wasn't a `/hold` label set. @dhaiducek are you also reviewing this? I can revert if you have comments

dhaiducek:
No worries--I'll take another look tomorrow, but quick look looks good to me! I especially like the `orderViaDependencies` to generate dependencies automatically!

I suppose the only thing I'm wondering about is whether `orderViaDependencies` should be moved to `policyDefaults` so that it can be specified at the individual policy level. It'd add another layer of complexity but might be nice to, for example, deploy an operator policy with all the operator things having dependencies and then have everything else not have dependencies?

dhaiducek:
Actually, the `extraDependencies` field is missing here. Are there future plans to add it?

JustinKuli:
Looking at this with fresher eyes, I'm confused by some of my choices - which might not be a good sign!

I think I incorrectly doc'd `orderViaDependencies`: it says "order they are defined in the manifests list" but actually it orders the output policies based on their order in the *policies* list. With that in mind, it could be useful to have a similar option in `policyDefaults` / individual policies to use `extraDependencies` to order the ConfigurationPolicies based on their order in the manifests list.

Maybe it looked hard when I was in the code implementing it, but now it definitely seems like adding `ignorePending` and `extraDependencies` options to the manifest options would be good.

@dhaiducek 
> I suppose the only thing I'm wondering about is whether orderViaDependencies should be moved to policyDefaults so that it can be specified at the individual policy level. It'd add another layer of complexity but might be nice to, for example, deploy an operator policy with all the operator things having dependencies and then have everything else not have dependencies?

Setting `orderViaDependencies` on individual policies seems hard to reason through: if I have 5 policies, and I set it on all except for the third, do I want a dependency graph like `1->2, 3->4->5` or `1->2, 3, 4->5` or `1->2->4->5, 3`? To prevent ambiguity, I'd rather just have it be all-or-nothing. If a user wants a policy outside of the chain, or working some other way, they can put it in a separate generator file!

dhaiducek:
Oh wait. I see what you're saying now. What about not allowing overrides at `policies[].orderViaDependencies` but allowing it within the policy for `extraDependencies`?

dhaiducek:
...or like you said (now that I'm reading it again), having an additional setting for the `extraDependencies` ordering, like `orderViaExtraDependencies` or something.

internal/plugin.go
```diff
@@ -513,6 +536,43 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
...
+			if dep.Kind == "" {
+				policy.Dependencies[i].Kind = policyKind
+			}
+
+			if dep.APIVersion == "" {
+				policy.Dependencies[i].APIVersion = policyAPIVersion
+			}
+		}
+
+		policy.Dependencies = append(policy.Dependencies, p.PolicyDefaults.Dependencies...)
```
mprahl, line 564: Everywhere else in the code, if you set it at a lower level, it replaces the value and doesn't act as a union. I would like the behavior to be consistent.

mprahl: This would also need to account if I explicitly set an empty array for `dependencies` at the policy level so the policy default is not inherited.

JustinKuli: Consistency would be good...

If the user sets an empty array for `dependencies` in a policy, *and* has `orderViaDependencies` set, how do you think those should interact? Right now, everything is merged so it's unambiguous (but inconsistent with other settings). I think I'd like the policy to still have the `order` dependency set - if I wanted a policy without that, I'd put it in a separate generator file.

dhaiducek: So it sounds like setting `dependencies` and setting `orderViaDependencies` should be mutually exclusive and throw an error? I think that's how those sorts of conflicts are handled elsewhere. I guess it's either that or have the ordering skip over policies that have `dependencies` specified in the ordering logic and continue with the next policy in the list?

mprahl: That's a good question. I lean towards `dependencies` and `orderViaDependencies` has being a union, but I would be okay with erroring if both are set.

PR #84, opened 2022-11-14
Policy extra ordering

JustinKuli:
Follow up to https://github.com/stolostron/policy-generator-plugin/pull/82

Don't be alarmed by the huge diff: most of it is tests, in a separate commit. The main impactful changes are in the "Second try implementation of policy ordering" commit, summary:
```
What was `orderViaDependencies` has been moved into `policyDefaults` as
`orderPolicies`. A similar `orderManifests` option will automatically
generate manifest-level `extraDependencies` to order ConfigPolicies
within a Policy. In both cases, instead of *merging* dependency lists,
it is only allowed to define dependencies one way.

If `dependencies` or `extraDependencies` are defined at multiple levels,
the more "specific" level will override previous definitions, which is
more consistent with other settings.
```


willkutler (approved):

willkutler:
This looks good to me! It will be very helpful to be able to automatically generate template-specific dependencies.

PR #85, opened 2022-11-22
Fix order manifests

JustinKuli:
Some fixes to the policy ordering feature after an end-to-end demo exposed some problems. More details in each commit.


mprahl:
@JustinKuli could you please fix the linting failure? Otherwise, this looks good.

mprahl (approved):

PR #87, opened 2022-12-02
Allow placement generation to be shut off

### LINKED ISSUE #86: Flag to disable placement rule generation
When a placement is not defined in the PolicyGenerator file an empty placement rule is created like so.

```
spec:
  clusterSelector:
    matchExpressions: []
```

This placement rule ends up targeting all managed clusters which is not the desired effect in some scenarios.

I propose a flag that can be set in the PolicyGenerator file that disables the placementRule creation.

dhaiducek:
When Policies are generated separately from PolicySets, it may be desirable to only generate the Policies and not the placement manifests.

Closes #86


JustinKuli (approved):
LGTM

PR #88, opened 2022-12-09
Fix go install and module path

cblecker:
It appears this is a fork of github.com/open-cluster-management-io/ocm-kustomize-generator-plugins, however the go module path was never set for this repo. This means that functionality like `go install github.com/stolostron/policy-generator-plugin/cmd/PolicyGenerator@v1.10.0` doesn't work.

This PR fixes this, allowing this functionality to work as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 *.so
 *.dylib
 **/.DS_Store
-PolicyGenerator
+/PolicyGenerator
 build_output
 
 # Test binary, built with `go test -c`
--- a/Makefile
+++ b/Makefile
@@ -48,11 +48,11 @@ clean:
 
 .PHONY: build
 build: layout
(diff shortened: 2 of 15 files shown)


dhaiducek:
Thanks for the PR, @cblecker! This is definitely an important usability update! The plan for this repo is to migrate it to the `open-cluster-management-io` org where it belongs and fork it back. (Originally we were asked to contribute it to the generic `ocm-kustomize-generator-plugins`, but since no other generators got added since then, we've gotten permission to have our repo in the org directly instead.) So once that happens, we'll have `open-cluster-management.io/policy-generator-plugin`.

dhaiducek:
@cblecker Sorry for the delay. Now that the generator is upstream, I've pulled these changes into a new PR, #107, and made associated updates. I'll keep an eye on it to verify, but this should resolve the `go install` command. Thanks for raising this PR!

PR #91, opened 2023-01-16
Add `policySetDefaults`

dhaiducek:
- Adds `policySetDefaults` to the generator
- Also moves placement validation and applying to helper functions for consistency and maintainability

ref:
- https://issues.redhat.com/browse/ACM-3003


internal/plugin.go
```diff
@@ -1068,262 +1062,210 @@ func (p *Plugin) assertValidConfig() error {
...
-			}
-		}
+	if placement.PlacementRuleName != "" && placement.PlacementName != "" {
+		return fmt.Errorf(
+			"%s must provide only one of placement.placementName or placement.placementRuleName", path,
+		)
 	}
 
-	seenPlcset := map[string]bool{}
+	if placement.PlacementRulePath != "" && placement.PlacementPath != "" {
```
mprahl, line 1154: This looks to be a duplicate check from above

internal/plugin.go
```diff
@@ -1068,262 +1062,210 @@ func (p *Plugin) assertValidConfig() error {
...
+		return fmt.Errorf(
+			"%s must provide only one of placement.placementPath or placement.placementRulePath", path,
+		)
+	}
 
-	for i := range p.PolicySets {
-		plcset := &p.PolicySets[i]
+	if len(placement.ClusterSelectors) > 0 && len(placement.LabelSelector) > 0 {
```
mprahl, line 1160: This looks to be a duplicate check from above

internal/plugin.go
```diff
@@ -1068,262 +1062,210 @@ func (p *Plugin) assertValidConfig() error {
...
+			"%s must provide only one of placement.labelSelector or placement.clusterSelectors", path,
+		)
+	}
 
-		if plcset.Name == "" {
-			return fmt.Errorf(
-				"each policySet must have a name set, but did not find a name at policySet array index %d", i,
-			)
-		}
+	if placement.PlacementRuleName != "" && placement.PlacementName != "" {
```
mprahl, line 1166: This looks to be a duplicate check from above

internal/plugin.go
dhaiducek: Ah, indeed--good catch!

internal/plugin.go
```diff
@@ -1209,121 +1101,153 @@ func (p *Plugin) assertValidConfig() error {
...
+		return fmt.Errorf(
+			"%s must provide only one of placement.placementName or placement.placementRuleName", path,
+		)
+	}
 
-		if plcset.Placement.PlacementRulePath != "" || plcset.Placement.PlacementPath != "" {
-			policySetPlacementOptions++
-		}
+	defaultPlacementOptions := 0
```
mprahl, line 1154: Maybe this would be better named as `placementOptions` since it's not strictly dealing with defaults.

internal/plugin.go
```diff
@@ -1209,121 +1101,153 @@ func (p *Plugin) assertValidConfig() error {
...
-		}
+	if placement.PlacementRulePath != "" || placement.PlacementPath != "" {
+		defaultPlacementOptions++
+	}
+
+	if placement.PlacementRuleName != "" || placement.PlacementName != "" {
+		defaultPlacementOptions++
+	}
+
+	if defaultPlacementOptions > 0 && !generatePlacement {
```
mprahl, line 1167: I don't necessarily think we should change this, but I was a bit surprised that I wouldn't be able to override a default placement from `policyDefaults` being applied to my policy/policy set by setting `generatePlacement: false` explicitly on it.

mprahl (approved):
Nice!

/hold in case you want to address the optional comments

internal/plugin.go
dhaiducek: Good point. I'll rename it to `placementOptionCount`. That seems even more accurate...

internal/plugin.go
dhaiducek: True--I suppose even if a placement is specified it becomes irrelevant when the `generatePlacement: false` flag is set. Especially in GitOps, it probably makes more sense here to skip this check and let things deploy. I'll take it out.

dhaiducek: Okay, now a placement can be specified but setting `generatePlacement` explicitly will take priority and not throw an error.

mprahl (approved):
/unhold

PR #93, opened 2023-01-26
Add support for generating placement with raw data

willkutler:
Placements can now be generated by nesting a full labelSelector or clusterSelector YAML under the `placement` key, in addition to the legacy method of listing key/value pairs under `placement`. This allows users to specify multiple clusters in the values array of a placement or placementRule, and to use matchLabels instead of matchExpressions (only supported by placement).

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -108,11 +108,15 @@ policyDefaults:
   # Optional. The placement configuration for the policies. This defaults to a placement configuration that matches all
   # clusters.
   placement:
-    # To specify a placement rule, specify key:value pair cluster selectors. (See placementRulePath to specify an
-    # existing file instead.)
+    # Deprecated: use clusterSelector instead
+    # To specify a placement rule, specify key:value pair cluster selectors or the full YAML for the desired cluster
+    # selectors. (See placementRulePath to specify an existing file instead.)
     clusterSelectors: {}
-    # To specify a placement, specify key:value pair cluster label selectors. (See placementPath to specify an existing
-    # file instead.)
+    # To specify a placement rule, specify key:value pair cluster selectors or the full YAML for the desired cluster
+    # selectors. (See placementRulePath to specify an existing file instead.)
+    clusterSelector: {}
+    # To specify a placement, specify key:value pair cluster label selectors or the full YAML for the desired cluster
+    # label selectors. (See placementPath to specify an existing file instead.)
     labelSelector: {}
     # Optional. Specifying a name will consolidate placement rules that contain the same cluster selectors.
     name: ""
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -3,6 +3,7 @@ package internal
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -12,7 +13,9 @@ import (
 	"time"
 
 	yaml "gopkg.in/yaml.v3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"open-cluster-management.io/ocm-kustomize-generator-plugins/internal/types"
 )
@@ -831,7 +834,7 @@ func (p *Plugin) assertValidConfig() error {
 	}
 
 	// Validate default policy placement settings
-	err := assertValidPlacement(p.PolicyDefaults.Placement, "policyDefaults", nil)
+	err := p.assertValidPlacement(p.PolicyDefaults.Placement, "policyDefaults", nil)
 	if err != nil {
 		return err
 	}
@@ -1061,14 +1064,14 @@ func (p *Plugin) assertValidConfig() error {
 			}
 		}
 
-		err := assertValidPlacement(policy.Placement, fmt.Sprintf("policy %s", policy.Name), &plCount)
+		err := p.assertValidPlacement(policy.Placement, fmt.Sprintf("policy %s", policy.Name), &plCount)
 		if err != nil {
 			return err
 		}
 	}
 
 	// Validate default policy set placement settings
-	err = assertValidPlacement(p.PolicySetDefaults.Placement, "policySetDefaults", nil)
+	err = p.assertValidPlacement(p.PolicySetDefaults.Placement, "policySetDefaults", nil)
 	if err != nil {
 		return err
 	}
@@ -1099,7 +1102,7 @@ func (p *Plugin) assertValidConfig() error {
 		seenPlcset[plcset.Name] = true
 
 		// Validate policy set Placement settings
-		err := assertValidPlacement(plcset.Placement, fmt.Sprintf("policySet %s", plcset.Name), &plCount)
+		err := p.assertValidPlacement(plcset.Placement, fmt.Sprintf("policySet %s", plcset.Name), &plCount)
 		if err != nil {
 			return err
 		}
@@ -1120,7 +1123,7 @@ func (p *Plugin) assertValidConfig() error {
 }
 
 // assertValidPlacement is a helper for assertValidConfig to verify placement configurations
-func assertValidPlacement(
+func (p *Plugin) assertValidPlacement(
 	placement types.PlacementConfig,
 	path string,
 	plCount *struct {
@@ -1235,6 +1238,26 @@ func assertValidPlacement(
 				)
 			}
 		}
+	}
+
+	if len(placement.ClusterSelectors) > 0 && len(placement.ClusterSelector) > 0 {
+		return fmt.Errorf("cannot use both clusterSelector and clusterSelectors in %s placement config "+
+			"(clusterSelector is recommended since it matches the actual placement field)", path)
+	}
+
+	// Determine which selectors to use
+	var resolvedSelectors map[string]interface{}
+	if len(placement.ClusterSelectors) > 0 {
+		resolvedSelectors = placement.ClusterSelectors
+	} else if len(placement.ClusterSelector) > 0 {
+		resolvedSelectors = placement.ClusterSelector
+	} else if len(placement.LabelSelector) > 0 {
+		resolvedSelectors = placement.LabelSelector
+	}
+
+	_, err := p.generateSelector(resolvedSelectors)
+	if err != nil {
+		return fmt.Errorf("%s placement has invalid selectors: %w", path, err)
 	}
 
 	return nil
@@ -1530,32 +1553,19 @@ func (p *Plugin) createPlacement(
 		}
 
 		// Determine which selectors to use
-		var resolvedSelectors map[string]string
+		var resolvedSelectors map[string]interface{}
 		if len(placementConfig.ClusterSelectors) > 0 {
 			resolvedSelectors = placementConfig.ClusterSelectors
+		} else if len(placementConfig.ClusterSelector) > 0 {
+			resolvedSelectors = placementConfig.ClusterSelector
 		} else if len(placementConfig.LabelSelector) > 0 {
 			resolvedSelectors = placementConfig.LabelSelector
 		}
 
-		// Sort the keys so that the match expressions can be ordered based on the label name
-		keys := make([]string, 0, len(resolvedSelectors))
-		for key := range resolvedSelectors {
-			keys = append(keys, key)
-		}
-		sort.Strings(keys)
-
-		matchExpressions := []map[string]interface{}{}
-		for _, label := range keys {
-			matchExpression := map[string]interface{}{
-				"key": label,
-			}
-			if resolvedSelectors[label] == "" {
-				matchExpression["operator"] = "Exists"
-			} else {
-				matchExpression["operator"] = "In"
-				matchExpression["values"] = []string{resolvedSelectors[label]}
-			}
-			matchExpressions = append(matchExpressions, matchExpression)
+		// Build cluster selector object
+		selectorObj, err := p.generateSelector(resolvedSelectors)
+		if err != nil {
+			return "", err
 		}
 
 		if p.usingPlR {
@@ -1567,9 +1577,7 @@ func (p *Plugin) createPlacement(
 					"namespace": p.PolicyDefaults.Namespace,
 				},
 				"spec": map[string]interface{}{
-					"clusterSelector": map[string]interface{}{
-						"matchExpressions": matchExpressions,
-					},
+					"clusterSelector": selectorObj,
 				},
 			}
 		} else {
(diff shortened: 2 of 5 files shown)


dhaiducek (changes requested):
Thanks! Some initial comments for consideration. Please also update the reference YAML to document these changes.

internal/plugin.go
```diff
@@ -1621,6 +1613,128 @@ func (p *Plugin) createPlacement(
...
+	for key := range resolvedSelectors {
+		if key == "matchExpressions" {
+			usedSelectors = append(usedSelectors, key)
+		} else if key == "matchLabels" {
+			usedSelectors = append(usedSelectors, key)
+		} else if !legacy {
+			usedSelectors = append(usedSelectors, "legacy")
+			legacy = true
+		}
+	}
```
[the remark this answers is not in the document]
mprahl: That doesn't work because the underlying type is always map[string]interface{}:
```go
package main

import "fmt"

func main() {
	temp := map[string]interface{}{"hello": "world"}
	var tempUntyped interface{} = temp
	_, ok := tempUntyped.(map[string]string)
	fmt.Printf("ok: %t\n", ok)
}
```

internal/plugin_test.go
```diff
@@ -3320,3 +3320,256 @@ func TestGenerateNonDNSBindingName(t *testing.T) {
...
+              operator: In
+              values:
+                - red hat
+                - test
+`
+	expected = strings.TrimPrefix(expected, "\n")
+	assertEqual(t, output, expected)
+}
+
+func TestCreatePlacementRuleFromMatchLabels(t *testing.T) {
```
mprahl, line 3422:
```suggestion
func TestCreatePlacementFromMatchLabels(t *testing.T) {
```

internal/plugin.go
```diff
@@ -1621,6 +1613,128 @@ func (p *Plugin) createPlacement(
...
+func (p *Plugin) generateSelector(
+	resolvedSelectors map[string]interface{},
+	keys []string,
+) (map[string]interface{}, error) {
+	// Determine cluster selector type
+	var selectorType string
+	var usedSelectors []string
+	var legacy bool
+
+	for key := range resolvedSelectors {
```
mprahl, line 1627: Could we rely on the typed structs in the apimachinery module? Here is a prototype of what I was thinking:

```go
func (p *Plugin) generateSelector(resolvedSelectors map[string]interface{}) (map[string]interface{}, error) {
	if resolvedSelectors == nil {
		return map[string]interface{}{"matchExpressions": []interface{}{}}, nil
	}

	resolvedSelectorsJSON, err := json.Marshal(resolvedSelectors)
	if err != nil {
		return nil, err
	}

	resolvedSelectorsLS := metav1.LabelSelector{}
	decoder := json.NewDecoder(bytes.NewReader(resolvedSelectorsJSON))
	decoder.DisallowUnknownFields()

	err = decoder.Decode(&resolvedSelectorsLS)
	if err != nil {
		resolvedSelectorsLS = metav1.LabelSelector{}

		// Check if it's a legacy selector
		for label, value := range resolvedSelectors {
			valueStr, ok := value.(string)
			if !ok {
				// TODO: Maybe an error message about the decode failing?
				return nil, fmt.Errorf(
					"legacy placement value has non-string type as value: %s",
					fmt.Sprint(resolvedSelectors[label]),
				)
			}

			lsReq := metav1.LabelSelectorRequirement{Key: label}

			if valueStr == "" {
				lsReq.Operator = metav1.LabelSelectorOpExists

			} else {
				lsReq.Operator = metav1.LabelSelectorOpIn
				lsReq.Values = []string{valueStr}
			}

			resolvedSelectorsLS.MatchExpressions = append(resolvedSelectorsLS.MatchExpressions, lsReq)
		}
	}

	if p.usingPlR && resolvedSelectorsLS.MatchLabels != nil {
		return nil, fmt.Errorf("placementRule does not support matchLabels")
	}

	resolved, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&resolvedSelectorsLS)
	if err != nil {
		// TODO: This should likely be a panic since it's a failure in generating the legacy label selector.
		panic(err)
	}

	return resolved, nil
}
```

internal/plugin.go
```diff
@@ -1621,6 +1613,128 @@ func (p *Plugin) createPlacement(
...
+
+	if len(usedSelectors) > 1 {
+		return nil, fmt.Errorf("too many placement selectors: %s", strings.Join(usedSelectors, ", "))
+	} else if len(usedSelectors) == 1 {
+		selectorType = usedSelectors[0]
+	}
+
+	if selectorType == "matchLabels" && p.usingPlR {
+		return nil, fmt.Errorf("placementRule does not support matchLabels")
+	}
```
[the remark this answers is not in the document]
willkutler: oops, I hadn't seen it used in a placementRule so I assumed it was only in Placement. Thanks

internal/plugin.go
willkutler: sure. I'll replace the selector generation logic with this

mprahl: @willkutler after thinking about this, if the JSON decode succeeds, maybe you can just return `resolvedSelectors` directly?

willkutler: @mprahl do you mean something like this? 
```
func (p *Plugin) generateSelector(
	resolvedSelectors map[string]interface{},
	keys []string,
) (map[string]interface{}, error) {
	if resolvedSelectors == nil {
		return map[string]interface{}{"matchExpressions": []interface{}{}}, nil
	}

	resolvedSelectorsJSON, err := json.Marshal(resolvedSelectors)
	if err != nil {
		return nil, err
	}

	resolvedSelectorsLS := metav1.LabelSelector{}
	decoder := json.NewDecoder(bytes.NewReader(resolvedSelectorsJSON))
	decoder.DisallowUnknownFields()

	err = decoder.Decode(&resolvedSelectorsLS)
	if err != nil {
		resolvedSelectorsLS = metav1.LabelSelector{}

		// Check if it's a legacy selector
		for label, value := range resolvedSelectors {
			valueStr, ok := value.(string)
			if !ok {
				return nil, fmt.Errorf(
					"matchExpressions has non-string type as value: %s",
					fmt.Sprint(resolvedSelectors[label]),
				)
			}

			lsReq := metav1.LabelSelectorRequirement{Key: label}

			if valueStr == "" {
				lsReq.Operator = metav1.LabelSelectorOpExists

			} else {
				lsReq.Operator = metav1.LabelSelectorOpIn
				lsReq.Values = []string{valueStr}
			}

			resolvedSelectorsLS.MatchExpressions = append(resolvedSelectorsLS.MatchExpressions, lsReq)
		}

		resolved, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&resolvedSelectorsLS)
		if err != nil {
			panic(err)
		}

		return resolved, nil
	}

	return resolvedSelectors, nil
}
```

docs/policygenerator-reference.yaml
```diff
@@ -108,11 +108,14 @@ policyDefaults:
   # Optional. The placement configuration for the policies. This defaults to a placement configuration that matches all
   # clusters.
   placement:
-    # To specify a placement rule, specify key:value pair cluster selectors. (See placementRulePath to specify an
-    # existing file instead.)
+    # To specify a placement rule, specify key:value pair cluster selectors or the full YAML for the desired cluster
+    # selectors. (See placementRulePath to specify an existing file instead.)
     clusterSelectors: {}
```
mprahl, line 113: Do you think we could just remove this from the docs so that new users start using `clusterSelector`?

docs/policygenerator-reference.yaml
```diff
@@ -108,11 +108,14 @@ policyDefaults:
...
-    # To specify a placement, specify key:value pair cluster label selectors. (See placementPath to specify an existing
-    # file instead.)
+    # To specify a placement rule, specify key:value pair cluster selectors or the full YAML for the desired cluster
+    # selectors. (See placementRulePath to specify an existing file instead.)
+    clusterSelector: {}
+    # To specify a placement, specify key:value pair cluster label selectors or the full YAML for the desired cluster
+    # label selectors. (See placementRulePath to specify an existing file instead.)
```
mprahl, line 118:
```suggestion
    # label selectors. (See placementPath to specify an existing file instead.)
```

internal/plugin.go
```diff
@@ -1237,6 +1240,40 @@ func assertValidPlacement(
 		}
 	}
 
+	if len(placement.ClusterSelectors) > 0 && len(placement.ClusterSelector) > 0 {
+		return fmt.Errorf("cannot use both clusterSelectors and clusterSelector in placement config")
```
mprahl, line 1244: Could you add a recommendation to use `clusterSelector`?

internal/plugin.go
```diff
@@ -1237,6 +1240,40 @@ func assertValidPlacement(
...
+	}
+
+	selectorsUsed := map[string]bool{
+		"legacy":       false,
+		"fullSelector": false,
+	}
+	for key := range placement.ClusterSelector {
```
mprahl, line 1251: I know it's a bit wasteful resource wise, but I think it'd be cleaner if you just ran `p.generatePlacement` and didn't use the output and just checked if the error was nil. All the validation is happening in that method already.

internal/plugin.go
```diff
@@ -1529,33 +1566,29 @@ func (p *Plugin) createPlacement(
 			return
 		}
 
+		if len(placementConfig.ClusterSelectors) > 0 && len(placementConfig.ClusterSelector) > 0 {
```
mprahl, line 1569: This shouldn't be required because you've validated this in `assertValidPlacement`.

docs/policygenerator-reference.yaml
willkutler: I think it might be a bit confusing to existing users if we completely remove it - what if we just added a recommendation to use clusterSelector here, similar to the error message?

mprahl: Sounds good!

willkutler:
@mprahl second round of review changes are in the latest commit

docs/policygenerator-reference.yaml
```diff
@@ -108,11 +108,14 @@ policyDefaults:
   # Optional. The placement configuration for the policies. This defaults to a placement configuration that matches all
   # clusters.
   placement:
-    # To specify a placement rule, specify key:value pair cluster selectors. (See placementRulePath to specify an
-    # existing file instead.)
+    # To specify a placement rule, specify key:value pair cluster selectors or the full YAML for the desired cluster
```
mprahl, line 111:
```suggestion
    # Deprecated: use clusterSelector instead
    # To specify a placement rule, specify key:value pair cluster selectors or the full YAML for the desired cluster
```

internal/plugin.go
```diff
@@ -1621,6 +1627,79 @@ func (p *Plugin) createPlacement(
...
+		if key == "matchExpressions" || key == "matchLabels" {
+			selectorsUsed["fullSelector"] = true
+		} else {
+			selectorsUsed["legacy"] = true
+		}
+	}
+
+	if selectorsUsed["legacy"] && selectorsUsed["fullSelector"] {
+		return nil, fmt.Errorf("cannot use both key/value pairs and formatted YAML for placement selectors")
+	}
```
mprahl, line 1655: Is this code needed since you detect if it's a legacy format in `for label, value := range resolvedSelectors {`?

internal/plugin.go
```diff
@@ -1621,6 +1627,79 @@ func (p *Plugin) createPlacement(
...
+	err = decoder.Decode(&resolvedSelectorsLS)
+	if err != nil {
+		resolvedSelectorsLS = metav1.LabelSelector{}
+
+		// Check if it's a legacy selector
+		for label, value := range resolvedSelectors {
+			valueStr, ok := value.(string)
+			if !ok {
+				return nil, fmt.Errorf(
+					"matchExpressions has non-string type as value: %s",
```
mprahl, line 1675: I think if you use this error then the validation code from above that I commented on can be deleted.
```suggestion
					"the input is not a valid label selector or simple key-value label matching map",
```

willkutler: @mprahl this is true but isn't it better to have the error message be a bit more specific?

mprahl: If I just put:
```yaml
clusterSelector:
    cloud: some-cloud
    somethingelse: 123
```

I'd be confused if the error said `matchExpressions`.

willkutler: ok I'll change it

willkutler:
@mprahl new review is done

docs/policygenerator-reference.yaml
mprahl: @willkutler could you please mention that `clusterSelectors` is deprecated?

internal/plugin.go
```diff
@@ -1237,6 +1240,26 @@ func assertValidPlacement(
...
+		resolvedSelectors = placement.ClusterSelectors
+	} else if len(placement.ClusterSelector) > 0 {
+		resolvedSelectors = placement.ClusterSelector
+	} else if len(placement.LabelSelector) > 0 {
+		resolvedSelectors = placement.LabelSelector
+	}
+
+	_, err := p.generateSelector(resolvedSelectors)
+	if err != nil {
+		return fmt.Errorf("invalid selectors in placement: %w", err)
```
mprahl, line 1260: Could you use the `path` method argument in the error message so that the user knows where the invalid selector is?

internal/plugin.go
```diff
@@ -1237,6 +1240,26 @@ func assertValidPlacement(
 		}
 	}
 
+	if len(placement.ClusterSelectors) > 0 && len(placement.ClusterSelector) > 0 {
+		return fmt.Errorf("cannot use both clusterSelector and clusterSelectors in placement config " +
```
mprahl, line 1244: Could you use the `path` method argument in the error message so that the user knows where the invalid selector is?

internal/plugin.go
```diff
@@ -1621,6 +1627,60 @@ func (p *Plugin) createPlacement(
...
+	err = decoder.Decode(&resolvedSelectorsLS)
+	if err != nil {
+		resolvedSelectorsLS = metav1.LabelSelector{}
+
+		// Check if it's a legacy selector
+		for label, value := range resolvedSelectors {
+			valueStr, ok := value.(string)
+			if !ok {
+				return nil, fmt.Errorf(
+					"the input is not a valid label selector or simple key-value label matching map",
```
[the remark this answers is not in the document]
willkutler: @mprahl I just went ahead and removed `simple`

internal/plugin.go
```diff
@@ -1237,6 +1240,26 @@ func assertValidPlacement(
 		}
 	}
 
+	if len(placement.ClusterSelectors) > 0 && len(placement.ClusterSelector) > 0 {
+		return fmt.Errorf("cannot use both clusterSelector and clusterSelectors in %s placement config"+
```
mprahl, line 1244:
```suggestion
		return fmt.Errorf("cannot use both clusterSelector and clusterSelectors in %s placement config "+
```

mprahl:
@willkutler this looks great. Just fix that one typo and squash your commits. I'll approve it afterwards.

willkutler:
@mprahl @dhaiducek I think this should be good to go

mprahl (approved):

PR #97, opened 2023-02-14
Change clusterSelector type to metav1.labelSelector allowing more option to define match expression

serngawy:


dhaiducek:
@serngawy Thanks for the PR, but providing a MatchExpression direcly is already implemented in:
- https://github.com/stolostron/policy-generator-plugin/pull/93

We did it this way using `map[string]interface{}` so that the old way (providing `key: value` pairs) would still be supported and it wouldn't be a breaking change.

Please try what you'd like to using a new version of the generator and let us know if there are gaps.

serngawy:
@dhaiducek Would you give example how to do it, I'm looking to have Value NotIn expression ?

dhaiducek:
It looks like there's a bug with the way we're handling the deprecated `placement.clusterSelectors`. It works to do:

```yaml
policyDefaults:
  namespace: policies
  placement:
    clusterSelectors:
      matchExpressions:
      - key: cloud
        operator: NotIn
        values:
        - red hat
```

Result:
```yaml
spec:
  clusterSelector:
    matchExpressions:
    - key: cloud
      operator: NotIn
      values:
      - red hat
```

But using `clusterSelector` results in an empty array:
```yaml
spec:
  clusterSelector:
    matchExpressions: []
```

I'll work on a fix.

PR #109, opened 2023-03-10
Implement `kustomizeOptions`

dhaiducek:
Allows passing flags equivalent to `--enable-alpha-plugins` and `--enable-helm`


mprahl:
@dhaiducek are there any security repercussions with this approach? I'm thinking in particular that when running in App Sub, the Helm execution would end up using App Sub's service account but don't know for sure.

dhaiducek:
> @dhaiducek are there any security repercussions with this approach? I'm thinking in particular that when running in App Sub, the Helm execution would end up using App Sub's service account but don't know for sure.

I was wondering about input on security implications. The risks feel similar to enabling Kustomize. I'm thinking it's toggled on by the user, and I think would call the internal Helm CLI, but that doesn't exist in App Sub, so it'd just fail in that case. (I just tried executing into the subscription container, and `helm` came up empty.)

That said, I suppose I'll have to do some more digging around ArgoCD, but the situation there might be similar.

eformat:
hi @dhaiducek .. this just "works" in argocd .. as it has helm and kustomize support already i.e. in the argocd-repo-server pod they have helm in /usr/local/bin/helm

mprahl:
@ch-stark and @dhaiducek , how about we require the user setting an environment variable to enable Helm in Kustomize? That way they opt-in to this feature and ensure it is secure to do so.

In Application Subscription, we can probably enable this by default and document it for ArgoCD.

dhaiducek:
> @ch-stark and @dhaiducek , how about we require the user setting an environment variable to enable Helm in Kustomize? That way they opt-in to this feature and ensure it is secure to do so.
> 
> In Application Subscription, we can probably enable this by default and document it for ArgoCD.

That works for me--I've been thinking we just need to get this feature out there, and having a higher-level opt in sounds like a good way to go.

mprahl (approved):

mprahl:
@dhaiducek could you please add documentation of the environment variable in the read me?

mprahl (approved):

dhaiducek:
Updating the `README.md` caused a really weird rebase (maybe because this PR was originally in `stolostron`?), but I think it's sorted now...

mprahl (approved):

PR #113, opened 2023-04-12
Wrap Gatekeeper manifests directly when expander is disabled

dhaiducek:
When `informGatekeeperPolicies` is set to `false`, wrap Gatekeeper manifests directly in a Policy rather than in a ConfigurationPolicy.

ref: https://issues.redhat.com/browse/ACM-4438

Additionally, I've noticed some cases where a Policy was being wrapped in a Policy by the generator, causing confusion. I added a commit here to throw an error in this case.

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -295,9 +295,16 @@ func isPolicyTypeManifest(manifest map[string]interface{}) (bool, error) {
 		return false, errors.New("invalid or not found kind")
 	}
 
-	isPolicy := strings.HasPrefix(apiVersion, "policy.open-cluster-management.io") &&
-		kind != "Policy" &&
-		strings.HasSuffix(kind, "Policy")
+	// Don't allow generation for root Policies
+	isOcmAPI := strings.HasPrefix(apiVersion, "policy.open-cluster-management.io")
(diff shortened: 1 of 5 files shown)


internal/utils.go
```diff
@@ -284,30 +287,43 @@ func setTemplateOptions(tmpl map[string]interface{}, ignorePending bool, extraDe
 // by checking apiVersion and kind fields.
 // Return error when apiVersion and kind fields aren't string, or if the manifest
 // is a non-root policy manifest, but it is invalid because it is missing a name.
-func isPolicyTypeManifest(manifest map[string]interface{}) (bool, error) {
+func isPolicyTypeManifest(manifest map[string]interface{}) (bool, bool, error) {
```
mprahl, line 290: Could you please update the doc for the function now that there's a new return value?

mprahl (dismissed):

internal/utils.go
dhaiducek: Good catch--done!

mprahl (dismissed):

dhaiducek:
Sorry for the churn--for some reason Tide couldn't rebase, so I rebased it myself.

mprahl (approved):

PR #120, opened 2023-07-14
Add version flag to generator

dhaiducek:
Adds a generator version flag with the Git SHA in the output:
```
v1.12.2-ab49eaf2a21d44265784bbd1f8381ec6e8928828
```

Also adds verification that the `VERSION.txt` file was updated so that a release isn't created without the version being updated.

This is my first time adding version information to anything, so if anyone has other ideas/pointers, do share!

ref: https://issues.redhat.com/browse/ACM-5803

Closes https://github.com/open-cluster-management-io/policy-generator-plugin/issues/118

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,14 +45,24 @@ clean:
 ############################################################
 # build section
 ############################################################
+GIT_VERSION := $(shell git describe --dirty 2>/dev/null)
+ifndef GIT_VERSION
+  GIT_BRANCH := $(shell git branch --show-current)
+  ifdef GIT_BRANCH
+    GIT_VERSION := $(GIT_BRANCH)-$(shell git rev-parse --short HEAD)
+  else
+    GIT_VERSION := $(shell git rev-parse --short HEAD)-dev
+  endif
+endif
+GO_LDFLAGS ?= -X 'main.Version=$(GIT_VERSION)'
 
 .PHONY: build
 build: layout
-	go build -o $(API_PLUGIN_PATH)/ ./cmd/PolicyGenerator
+	go build -ldflags="$(GO_LDFLAGS)" -o $(API_PLUGIN_PATH)/ ./cmd/PolicyGenerator
 
 .PHONY: build-binary
 build-binary:
-	go build ./cmd/PolicyGenerator
+	go build -ldflags="$(GO_LDFLAGS)" ./cmd/PolicyGenerator
 
 .PHONY: build-release
 build-release:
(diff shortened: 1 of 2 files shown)


cmd/PolicyGenerator/main.go
```diff
@@ -2,20 +2,48 @@ package main
...
 	"github.com/spf13/pflag"
 	"open-cluster-management.io/policy-generator-plugin/internal"
 )
 
+//go:embed VERSION.txt
+var version []byte
 var debug = false
 
+// Gather the Git commit
+var gitCommit = func() string {
```
mprahl, line 19:
```suggestion
func gitCommit() string {
```

cmd/PolicyGenerator/main.go
```diff
@@ -2,20 +2,48 @@ package main
...
+}()
+
 func main() {
 	// Parse command input
 	debugFlag := pflag.Bool("debug", false, "Print the stack trace with error messages")
+	versionFlag := pflag.Bool("version", false, "Print the version of the generator")
 	pflag.Parse()
 
+	if *versionFlag {
+		versionStr := string(version)
```
mprahl, line 38: Could you strip any new lines on this? I could see `VERSION.txt` have one by accident.

cmd/PolicyGenerator/VERSION.txt
```diff
@@ -0,0 +1 @@
+v1.12.2
```
mprahl, line 1: What is the advantage of a separate file as opposed to defining the value in a constant variable in `main.go`?

dhaiducek: Mostly copy/paste, but it is easier to parse externally (like in the GitHub Action check). I'm not attached to it if you think a constant would be the way to go!

mprahl: I don't have a strong preference but one option is that we could define a constant like `const version = "VERSION"` and then in the GitHub release workflow, there could be a sed command run to replace `"VERSION"` with the GitHub tag.

That way we wouldn't have to update the version manually.

dhaiducek: I definitely like not having to update the version manually! And we could use similar logic in the subscription image build to inject a version.

cmd/PolicyGenerator/main.go
```diff
@@ -2,20 +2,48 @@ package main
...
 	"fmt"
 	"os"
+	runtimeDebug "runtime/debug"
 
 	"github.com/spf13/pflag"
 	"open-cluster-management.io/policy-generator-plugin/internal"
 )
 
+//go:embed VERSION.txt
+var version []byte
```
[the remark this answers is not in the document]
dhaiducek: I really like this (and didn't know `git describe` existed!), except that `stolostron` uses branches instead of tags, so `git describe --dirty` doesn't return anything. But there could be a function in the Makefile to fill it in using the branch/commit if it's empty...

dhaiducek:
I've used @JustinKuli's idea in the latest code to default with fallbacks in the following order:
1. Native `git describe` output when available (`v1.12.1-3-gaa03f1c`)
2. `<branch>-<Git SHA>` when tags aren't available like in Stolostron (`release-2.9-8466f8c`)
3. `<git SHA>-dev` when a branch isn't available like a detached HEAD (`8466f8c-dev`)

We could maybe add the org in there, but this ought to be enough information/fallbacks to find the version at play (or figure out that it wasn't built by us).

Makefile
```diff
@@ -45,14 +45,24 @@ clean:
 ############################################################
 # build section
 ############################################################
+GIT_VERSION := $(shell git describe --dirty 2>/dev/null)
```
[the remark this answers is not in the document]
dhaiducek: More of a safety/notification thing than anything. Without it, the `-dirty` suffix isn't added when there have been local changes:
> When the working tree matches HEAD, the output is the same as "git describe HEAD". If the working tree has local modification "-dirty" is appended to it.

https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---dirtyltmarkgt

yiraeChristineKim (approved):

PR #126, opened 2023-09-06
Fix ExtraDependencies for ConfigPolicy manifests

dhaiducek:
When a `ConfigurationPolicy` manifest was provided directly, `extraDependencies` was being set at the `objectDefinition` level rather than the `policy-templates` level.

ref: https://issues.redhat.com/browse/ACM-7353


mprahl (approved):

internal/utils.go
```diff
@@ -189,7 +189,7 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 
 				// Only set dependency options if it's an OCM policy
 				if isOcmPolicy {
-					setTemplateOptions(manifest, ignorePending, extraDeps)
+					setTemplateOptions(policyTemplate, ignorePending, extraDeps)
```
JustinKuli, line 192: So, it was a one word change? Awesome! 😆

PR #128, opened 2023-09-08
ACM-5068: Set consolidateManifests=false when orderManifests=true

JeffeyL:
Description: (Copied from issue) When using the Policy Generator, when `orderManifests` is `true`, it's a signal that the user wants to have dependencies set for the manifests within. However, `consolidateManifests` defaults to `true`, so `orderManifests` won't really do anything unless `consolidateManifests` is also set to false. As a convenience, we should set `consolidateManifests` accordingly so that this implementation detail is abstracted for the user.

Solution: Conditionals were added where `consolidateManifests` was being set (both default and actual values) to set to `false` when the corresponding `orderManifests` value is `true`.

Ref: https://issues.redhat.com/browse/ACM-5068


JeffeyL:
There were some unrelated unit tests failing when I was testing, will check back.

internal/ordering_test.go
```diff
@@ -492,7 +492,7 @@ policies:
 			wantFile: "",
 			wantErr:  "extraDependencies may not be set in policy one manifest[0] because orderManifests is set",
 		},
-		"orderManifests and consolidateManifests is false": {
+		"orderManifests is true and consolidateManifests is unset": {
```
JustinKuli, line 495: Nit: can we add a tiny bit more context to these tests?
```suggestion
		"orderManifests is true in policyDefaults, and consolidateManifests is unset": {
```

internal/ordering_test.go
```diff
@@ -509,8 +509,8 @@ policies:
   - path: {{printf "%v/%v" .Dir "configmap.yaml"}}
   - path: {{printf "%v/%v" .Dir "configmap.yaml"}}
 `,
-			wantFile: "",
-			wantErr:  "policyDefaults may not specify both consolidateManifests and orderManifests",
+			wantFile: "testdata/ordering/three-ordered-manifests.yaml",
+			wantErr:  "",
 		},
 		"orderManifests in policy and consolidateManifests is false": {
```
JustinKuli, line 515: If we can update this description as well, something like
```suggestion
		"orderManifests is true in policy and consolidateManifests is unset": {
```

JustinKuli (commented):
Looks good overall to me!

PR #132, opened 2023-11-16
Set the severity annotation for non-OCM policies

mprahl:
For policy-templates entries that are not OCM policies, it is supported to set the `policy.open-cluster-management.io/severity` annotation for other tooling to determine the severity of a noncompliant event. This adds that automatically.

Relates:
https://issues.redhat.com/browse/ACM-6860


dhaiducek (changes requested):
Gatekeeper is our current use-case, but I think this should be more generic, removing the fact that it's Gatekeeper and just setting the annotation if it's not OCM? It'd make the logic simpler and also prevent reverting this change if we ever add integrations for other policy engines. What do you think?

dhaiducek (approved):
LGTM! Thanks! 🙂

PR #135, opened 2023-12-08
Make `Placement` the default

dhaiducek:
`PlacementRule` is deprecated

ref: https://issues.redhat.com/browse/ACM-1286


dhaiducek:
That's more like it 🙂

dhaiducek:
The ManagedClusterSetBinding definitely puts a wrinkle here as far as UX goes, but we'll have to do this eventually...

mprahl:
@dhaiducek , I think this will need a rebase after #136

PR #137, opened 2023-12-18
Fix setting manifest configurations

dhaiducek:
- Fix severity set at the manifest level
- Fix setting manifest overrides correctly
  -  _Since the manifest configuration wasn't being set when consolidateManifests was true, it wasn't being propagated to things like expanded policies and policy template manifests provided directly. This fixes it to set manifest configurations and then compare manifest and policy configurations to make sure they match when consolidateManifests is true._

ref: https://issues.redhat.com/browse/ACM-9131


dhaiducek:
Didn't account for Placement changes in `main`. Updating.

mprahl (approved):

PR #139, opened 2023-12-19
Allow setting policy open-cluster-management.io/description annoation

brian-jarvis:
Adds description field to PolicyOptions.  And sets the default value if not defined in the PolicyGenerator.

https://issues.redhat.com/browse/ACM-9080


dhaiducek (approved):
LGTM! Thanks for filling in this gap in the generator!

PR #148, opened 2024-02-09
Set unreachable Toleration in Placement

dhaiducek:
Without it, ManagedClusters that are unreachable are patched with a Taint and aren't returned by the Placement controller.

(Making `Placement` more widely configurable, i.e. adding `tolerations`, is outside the scope of this PR since it would require refactoring or adding an additional `placement` configuration alongside `labelSelector`.)

ref: https://issues.redhat.com/browse/ACM-9871


internal/plugin.go
```diff
@@ -1652,6 +1652,12 @@ func (p *Plugin) createPlacement(
 							},
 						},
 					},
+					"tolerations": []map[string]interface{}{
+						{
+							"key": "cluster.open-cluster-management.io/unreachable",
+							"operator": "Exists",
+						},
+					},
```
dhaiducek, line 1660: This is the change--the rest of the PR are test updates.

yiraeChristineKim (approved):

mprahl (approved):

PR #154, opened 2024-03-26
Fix PolicyGenerator setting InformOnly on root policy

JeffeyL:
When all ConfigurationPolicies have `RemediationAction` `InformOnly`, the root Policy is also set to have `RemediationAction` `InformOnly`. However, `InformOnly` is not valid on a Policy, and was thus modified to be set to `inform` in this case.

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -598,5 +598,10 @@ func getRootRemediationAction(policyTemplates []map[string]interface{}) string {
 		}
 	}
 
+	// "InformOnly" should only apply to ConfigurationPolicies
+	if strings.EqualFold(action, "informonly") {
+		action = "inform"
+	}
+
 	return action
 }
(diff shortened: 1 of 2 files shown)


mprahl (approved):

internal/utils.go
```diff
@@ -598,5 +598,10 @@ func getRootRemediationAction(policyTemplates []map[string]interface{}) string {
 		}
 	}
 
+	// "InformOnly" should only apply to ConfigurationPolicies
+	if action == "InformOnly" {
```
mprahl, line 602: It seems we accept any casing so this should also accept that too:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/6a8baf33b678533fb6ebe268edd47a421cc81143/controllers/templatesync/template_sync.go#L1345

mprahl (changes requested):
/unhold

mprahl (approved):

PR #157, opened 2024-04-03
Allow override across different placement types

zyjjay:
This change will allow placements to override placementrule defaults and vice versa.

ref: https://issues.redhat.com/browse/ACM-10152

### DIFF
--- a/internal/plugin_config_test.go
+++ b/internal/plugin_config_test.go
@@ -788,13 +788,9 @@ policies:
 	p := Plugin{}
 
 	err = p.Config([]byte(config), tmpDir)
-	if err == nil {
-		t.Fatal("Expected an error but did not get one")
+	if err != nil {
+		t.Fatal(err.Error())
 	}
-
-	expected := "policy policy-app-config must specify only one of " +
-		"placement selector, placement path, or placement name"
-	assertEqual(t, err.Error(), expected)
 }
 
 func TestConfigMultipleDefaultAndPolicyPlacementNames(t *testing.T) {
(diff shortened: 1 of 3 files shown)


dhaiducek:
> This change will allow placements to override placementrule defaults and vice versa.

Sorry, it looks like the issue was worded poorly. I don't think we should allow overriding `Placement` for `PlacementRule`, especially when `PlacementRule` is deprecated. However, when considering the options for a particular placement Kind (file path, selector, or name), these should be override-able if provided in the `policies` array.

zyjjay:
So based on my understanding, only placement Kinds (name, fp, selector) in the policies array should override the default placement/placementRule Kinds. If a user were to specify a placementRule Kind in a policy in the policies array, but the policy default specifies a placement Kind, would the generated policy bind to a placement or a placementRule?

dhaiducek:
Sorry, you have it right--I think I got my wires crossed. Let me give this PR another look.

zyjjay:
I currently have it so that a placement Kind could override a placementRule default, and vice versa. I could modify it so that **only** placement Kinds in the policies array could override placementRule defaults if that's the desired behavior.

dhaiducek:
No, I think it's correct to have it this way. Sorry for my confusion. 😅

internal/plugin_config_test.go
```diff
@@ -788,13 +788,9 @@ policies:
 	p := Plugin{}
 
 	err = p.Config([]byte(config), tmpDir)
-	if err == nil {
-		t.Fatal("Expected an error but did not get one")
+	if err != nil {
+		t.Fatal("Did not expect errors but got one")
```
dhaiducek, line 792: This should also return the error message now that it's no longer `nil`.

internal/plugin_config_test.go
```diff
@@ -824,13 +820,9 @@ policies:
 	p := Plugin{}
 
 	err := p.Config([]byte(config), tmpDir)
-	if err == nil {
-		t.Fatal("Expected an error but did not get one")
+	if err != nil {
+		t.Fatal("Did not expect errors but got one")
```
dhaiducek, line 824: This should also return the error message now that it's no longer `nil`.

dhaiducek (changes requested):
Minor testing comments, but overall looks good! Thanks for updating it (and straightening me out!).

(Optional) I think you could add one more test to test overriding PlacementRule with Placement?

dhaiducek (approved):
LGTM! Thanks for updating it!

PR #158, opened 2024-04-11
Add support for object-templates-raw

JeffeyL:
Adds support for manifest files with only object-templates-raw field, which gets put into a ConfigurationPolicy.

Ref: https://issues.redhat.com/browse/ACM-10565

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -216,7 +216,10 @@ policies:
       #   1) Non-root policy type manifests such as IamPolicy, CertificatePolicy, and ConfigurationPolicy that have a
       #      "Policy" suffix. These are not modified except for patches and are directly added as a Policy's
       #      policy-templates entry.
-      #   2) For everything else, ConfigurationPolicy objects are generated to wrap these manifests. The resulting
+      #   2) Manifests containing only an `object-templates-raw` key. The corresponding value will be used directly in
+      #      a generated ConfigurationPolicy without modification, which will then be added as a Policy's 
+      #      policy-templates entry.
+      #   3) For everything else, ConfigurationPolicy objects are generated to wrap these manifests. The resulting
       #      ConfigurationPolicy is added as a Policy's policy-templates entry.
       - path: ""
         # Optional. (See policyDefaults.complianceType for description.)
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -187,7 +187,19 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 			}
 
 			if isPolicyTypeManifest {
-				policyTemplate := map[string]interface{}{"objectDefinition": manifest}
+				var policyTemplate map[string]interface{}
+
+				_, found, _ := unstructured.NestedString(manifest, "object-templates-raw")
+				if found {
+					policyTemplate = buildPolicyTemplate(
+						policyConf,
+						len(policyTemplates)+1,
+						manifest["object-templates-raw"],
+						&policyConf.Manifests[i].ConfigurationPolicyOptions,
+					)
+				} else {
+					policyTemplate = map[string]interface{}{"objectDefinition": manifest}
+				}
 
 				// Only set dependency options if it's an OCM policy
 				if isOcmPolicy {
@@ -309,6 +321,13 @@ func setTemplateOptions(tmpl map[string]interface{}, ignorePending bool, extraDe
 // - the manifest is a root policy manifest
 // - the manifest is invalid because it is missing a name
 func isPolicyTypeManifest(manifest map[string]interface{}, informGatekeeperPolicies bool) (bool, bool, error) {
+	// check for object-templates-raw separate from policies since they have separate requirements
+	_, found, _ := unstructured.NestedString(manifest, "object-templates-raw")
+	if found {
+		// return true for isPolicyType, since object-templates-raw is in a ConfigurationPolicy
+		return true, true, nil
+	}
+
 	apiVersion, found, err := unstructured.NestedString(manifest, "apiVersion")
 	if !found || err != nil {
 		return false, false, errors.New("invalid or not found apiVersion")
@@ -398,7 +417,7 @@ func processKustomizeDir(path string) ([]map[string]interface{}, error) {
 func buildPolicyTemplate(
 	policyConf *types.PolicyConfig,
 	policyNum int,
-	objectTemplates []map[string]interface{},
+	objectTemplates interface{},
 	configPolicyOptionsOverrides *types.ConfigurationPolicyOptions,
 ) map[string]interface{} {
 	var name string
@@ -408,6 +427,18 @@ func buildPolicyTemplate(
 		name = policyConf.Name
 	}
 
+	policySpec := map[string]interface{}{
+		"remediationAction": policyConf.RemediationAction,
+		"severity":          policyConf.Severity,
+	}
+
+	switch objectTemplates.(type) {
+	case string:
+		policySpec["object-templates-raw"] = objectTemplates
+	case []map[string]interface{}:
+		policySpec["object-templates"] = objectTemplates
+	}
+
 	policyTemplate := map[string]interface{}{
 		"objectDefinition": map[string]interface{}{
 			"apiVersion": policyAPIVersion,
@@ -415,11 +446,7 @@ func buildPolicyTemplate(
 			"metadata": map[string]interface{}{
 				"name": name,
 			},
-			"spec": map[string]interface{}{
-				"object-templates":  objectTemplates,
-				"remediationAction": policyConf.RemediationAction,
-				"severity":          policyConf.Severity,
-			},
+			"spec": policySpec,
 		},
 	}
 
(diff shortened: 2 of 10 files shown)


internal/utils.go
```diff
@@ -189,6 +190,39 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
...
+
+					policyTemplate = map[string]interface{}{
+						"objectDefinition": map[string]interface{}{
+							"apiVersion": policyAPIVersion,
+							"kind":       configPolicyKind,
+							"metadata": map[string]interface{}{
+								"name": name,
+							},
+							"spec": map[string]interface{}{
+								"object-templates-raw": objectTemplatesRaw,
```
dhaiducek, line 212: I think doing it this way might miss some configurations, like dependencies and diff logging. Rather than handling `object-templates-raw` as a policy type, could we instead somehow pass it through to `buildPolicyTemplates()` directly and inject it there?
https://github.com/open-cluster-management-io/policy-generator-plugin/blob/a097bd7881bb6ab72032449e316b1c691150f094/internal/utils.go#L419

There might be other/better ways, but I'm thinking you could rename the `objectTemplates` argument to `policySpec` and set either the `object-templates-raw` or `object-templates` key directly as part of it. Then set remediationAction and severity before setting the `spec` directly:
```golang
policySpec["remediationAction"] = policyConf.RemediationAction
policySpec["severity"] = policyConf.Severity
...
"spec": policySpec,
```

JeffeyL: Thanks, I'll look into it.

internal/utils.go
```diff
@@ -408,18 +425,29 @@ func buildPolicyTemplate(
...
+
+	objectTemplatesRawContent, exists := policySpec[0]["object-templates-raw"]
+	if exists {
+		policyTemplateSpec = map[string]interface{}{
+			"object-templates-raw": objectTemplatesRawContent,
+			"remediationAction":    policyConf.RemediationAction,
+			"severity":             policyConf.Severity,
+		}
+	}
+
```
dhaiducek, line 442: What you have looks good, but I keep rethinking this. Building on what you have, perhaps if we pass in an `interface{}` for `objectTemplates` and then parse the type to set the key accordingly? Something like:

```golang
	policySpec := map[string]interface{}{}
	policySpec["remediationAction"] = policyConf.RemediationAction
	policySpec["severity"] = policyConf.Severity

	if _, ok := objectTemplates.(string); ok {
		policySpec["object-templates-raw"] = objectTemplates
	} else if _, ok := objectTemplates.([]map[string]interface{}); ok {
		policySpec["object-templates"] = objectTemplates
	}
```

JeffeyL: That does seem like it would be cleaner!

internal/utils.go
```diff
@@ -309,6 +319,13 @@ func setTemplateOptions(tmpl map[string]interface{}, ignorePending bool, extraDe
 // - the manifest is a root policy manifest
 // - the manifest is invalid because it is missing a name
 func isPolicyTypeManifest(manifest map[string]interface{}, informGatekeeperPolicies bool) (bool, bool, error) {
+	// check for object-templates-raw separate from policies since they have separate requirements
+	_, found, _ := unstructured.NestedString(manifest, "object-templates-raw")
+	if found {
+		// return true for isPolicyType, since object-templates-raw is in a ConfigurationPolicy
+		return true, false, nil
```
dhaiducek, line 326: The second boolean, whether it's an OCM policy, should also be `true` so that it gets passed through the `setTemplateOptions()` function.
```suggestion
		return true, true, nil
```

dhaiducek (changes requested):
Looking good!

Unless I missed it, could you also add some additional parameters to the tests, like extraDependencies, ignorePending, or evaluationInterval, to make sure those are getting set properly?

internal/utils.go
```diff
@@ -189,6 +189,16 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
...
 
+				_, found, _ := unstructured.NestedString(manifest, "object-templates-raw")
+				if found {
+					policyTemplate = buildPolicyTemplate(
+						policyConf,
+						len(policyTemplates)+1,
+						manifest["object-templates-raw"],
+						&policyConf.Manifests[i].ConfigurationPolicyOptions,
+					)
+				}
```
mprahl, line 200:
```suggestion
				var policyTemplate map[string]interface{}

				_, found, _ := unstructured.NestedString(manifest, "object-templates-raw")
				if found {
					policyTemplate = buildPolicyTemplate(
						policyConf,
						len(policyTemplates)+1,
						manifest["object-templates-raw"],
						&policyConf.Manifests[i].ConfigurationPolicyOptions,
					)
				} else {
						policyTemplate = map[string]interface{}{"objectDefinition": manifest}
				}
```

internal/utils_test.go
```diff
@@ -1262,6 +1262,74 @@ func TestGetPolicyTemplateInvalidManifest(t *testing.T) {
 	assertEqual(t, err.Error(), expected)
 }
 
+func TestGetPolicyTemplateObjectTemplatesRaw(t *testing.T) {
+	t.Parallel()
+	tmpDir := t.TempDir()
+	manifestPath := path.Join(tmpDir, "configmap.yaml")
+	manifestYAML := `
+object-templates-raw: |
```
mprahl, line 1270: Could you have a test to verify when a single manifest file has two object-templates-raw fields? For example:

```yaml
object-templates-raw: |
  something
----
object-templates-raw: |
  something2
```

JeffeyL: I ended up just making the test have two `object-templates-raw` fields instead of one. Would it still be necessary to test the singular case?

mprahl: I think one test with both in it is fine.

internal/utils_test.go
```diff
@@ -1262,6 +1262,76 @@ func TestGetPolicyTemplateInvalidManifest(t *testing.T) {
...
+	}
+
+	objectTemplatesRaw, ok := spec["object-templates-raw"].(string)
+	if !ok {
+		t.Fatal("The object-templates-raw field is an invalid format")
+	}
+
+	assertEqual(t, len(objectTemplatesRaw), len(manifestYAMLContent))
+
+	policyTemplate2 := policyTemplates[0]
```
mprahl, line 1319:
```suggestion
	policyTemplate2 := policyTemplates[1]
```

internal/utils_test.go
```diff
@@ -1262,6 +1262,76 @@ func TestGetPolicyTemplateInvalidManifest(t *testing.T) {
...
+	if !ok {
+		t.Fatal("The spec field is an invalid format")
+	}
+
+	objectTemplatesRaw, ok := spec["object-templates-raw"].(string)
+	if !ok {
+		t.Fatal("The object-templates-raw field is an invalid format")
+	}
+
+	assertEqual(t, len(objectTemplatesRaw), len(manifestYAMLContent))
```
mprahl, line 1317: Can you check the actual value instead of length?

internal/utils_test.go
```diff
@@ -1262,6 +1262,76 @@ func TestGetPolicyTemplateInvalidManifest(t *testing.T) {
...
+	if !ok {
+		t.Fatal("The spec field is an invalid format")
+	}
+
+	objectTemplatesRaw, ok = spec["object-templates-raw"].(string)
+	if !ok {
+		t.Fatal("The object-templates-raw field is an invalid format")
+	}
+
+	assertEqual(t, len(objectTemplatesRaw), len(manifestYAMLContent))
```
mprahl, line 1332: Can you check the actual value instead of length?

mprahl (dismissed):
Nice!

/hold in case @dhaiducek wants to take another look.

dhaiducek (changes requested):
One more thing I thought of: for due diligence, please also add one configuration to the `examples/` directory and also add to the `manifests` description that we've enabled this:
https://github.com/open-cluster-management-io/policy-generator-plugin/blob/a097bd7881bb6ab72032449e316b1c691150f094/docs/policygenerator-reference.yaml#L215

docs/policygenerator-reference.yaml
```diff
@@ -216,7 +216,10 @@ policies:
       #   1) Non-root policy type manifests such as IamPolicy, CertificatePolicy, and ConfigurationPolicy that have a
       #      "Policy" suffix. These are not modified except for patches and are directly added as a Policy's
       #      policy-templates entry.
-      #   2) For everything else, ConfigurationPolicy objects are generated to wrap these manifests. The resulting
+      #   2) Manifests containing only a `object-templates-raw` key. The corresponding value will be used directly in
```
mprahl, line 219:
```suggestion
      #   2) Manifests containing only an `object-templates-raw` key. The corresponding value will be used directly in
```

mprahl (approved):

mprahl:
@dhaiducek let me know if this is good from your perspective to unhold.

dhaiducek (approved):
LGTM! Thanks!
The `object-templates-raw` example doesn't actually have any templates in it, but I think that's fine--if anyone comes looking around in the generator, they're probably already familiar with ConfigurationPolicy anyway.

PR #165, opened 2024-06-21
Add support for disabling load restrictions for executing kustomize

mprahl:
The user may set POLICY_GEN_DISABLE_LOAD_RESTRICTORS=true to disable the Kustomize load restrictions when the manifest is a Kustomize directory. This is off by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,8 @@ make build
    ```
 
   **NOTE:** To enable Helm processing when passing a Kustomize directory into the generator, set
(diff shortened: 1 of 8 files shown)


dhaiducek (approved):
LGTM!

If I recall, this doesn't lift any of our own path parsing restrictions, right? Only for Helm?

mprahl:
> LGTM!
> 
> If I recall, this doesn't lift any of our own path parsing restrictions, right? Only for Helm?

Right, this option only applies if the user specifies a Kustomize directory as a manifest.

PR #167, opened 2024-07-09
Add new field gatekeeperEnforcementAction to override enforcementAction

yiraeChristineKim:
- Add a new field of `gatekeeperEnforcementAction`
- Default value of the field is unset and when unset, don't touch the constraint (existing behavior) 
- If set, then directly set `spec.enforcementAction` in the Gatekeeper constraint 
- Should be settable at the policyDefaults, policy, and manifest levels (i.e. like the `ConfigurationPolicyOptions` struct).
Ref: https://issues.redhat.com/browse/ACM-11076
Signed-off-by: yiraeChristineKim <yikim@redhat.com>

### DIFF
--- a/docs/policygenerator-reference.yaml
+++ b/docs/policygenerator-reference.yaml
@@ -199,7 +199,10 @@ policyDefaults:
   # Optional. Labels that the policy will include under its metadata.labels. It will be applied for all
   # policies unless specified in the policy.
   policyLabels: {}
-
+  # Optional. Overrides the spec.enforcementAction field of a Gatekeeper constraint. 
+  # This only applies to Gatekeeper constraints and is ignored by other manifests. 
+  # If not set, the spec.enforcementAction field is not changed.
+  gatekeeperEnforcementAction: "deny"
 # Optional. Defaults for policy set generation. Any default value listed here can be overridden under an entry in the
 # policySets array.
 policySetDefaults:
@@ -260,6 +263,8 @@ policies:
         # Optional. (See policyDefaults.severity for description.)
         # Cannot be specified when policyDefaults.consolidateManifests is set to true.
         severity: "low"
+        # Optional. (See policyDefaults.gatekeeperEnforcementAction for description.)
+        gatekeeperEnforcementAction: "warn"
         # (Note: a path to a directory containing a Kustomize manifest is a supported alternative.) Optional. A
         # Kustomize patch to apply to the manifest(s) at the path. If there are multiple manifests, the patch requires
         # the apiVersion, kind, metadata.name, and metadata.namespace (if applicable) fields to be set so Kustomize can
@@ -351,6 +356,8 @@ policies:
     policyAnnotations: {}
     # Optional. (See policyDefaults.policyLabels for description.)
     policyLabels: {}
+    # Optional. (See policyDefaults.gatekeeperEnforcementAction for description.)
+    gatekeeperEnforcementAction: "dryrun"
 
 # Optional. The list of policy sets to create. To include a policy in a policy set, use policies[*].policySets or
 # policyDefaults.policySets or policySets.policies.
--- a/internal/plugin_test.go
+++ b/internal/plugin_test.go
@@ -261,6 +261,7 @@ metadata:
 policyDefaults:
   namespace: my-policies
   consolidateManifests: false
+  gatekeeperEnforcementAction: deny
   %s: %s
 policies:
 - name: policy-app
@@ -281,6 +282,7 @@ policies:
 					t.Fatal("Unexpected error", err)
 				}
 
+				assertEqual(t, p.PolicyDefaults.GatekeeperEnforcementAction, "deny")
 				assertEqual(t, p.Policies[0].ConsolidateManifests, false)
 
 				output, err := p.Generate()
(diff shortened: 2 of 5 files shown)


docs/policygenerator-reference.yaml
```diff
@@ -199,7 +199,10 @@ policyDefaults:
   # Optional. Labels that the policy will include under its metadata.labels. It will be applied for all
   # policies unless specified in the policy.
   policyLabels: {}
-
+  # Optional. gatekeeperEnforcementAction("deny" | "warn" | "dryrun") will override gatekeeper.constraint.spec.enforcementAction
+  # This only applies to gatekeeper constraint
+  # The default is "deny"
```
mprahl, line 204:
```suggestion
  # Optional. Overrides the spec.enforcementAction field of a Gatekeeper constraint. This only applies to Gatekeeper constraints and is ignored by other manifests. If not set, the spec.enforcementAction field is not changed.
```

internal/utils.go
```diff
@@ -217,6 +217,12 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 					annotations[severityAnnotation] = policyConf.Severity
 
 					policyTemplateUnstructured.SetAnnotations(annotations)
+
+					err := setGatekeeperEnforcementAction(policyTemplate,
```
mprahl, line 221: It might be cleaner to directly work with the manifest rather than after it is a policy template. You could do this right at the start of the `for _, manifest := range manifestGroup {` loop.

internal/plugin_test.go
```diff
@@ -1760,6 +1762,162 @@ spec:
...
+		policyConf      types.PolicyConfig
+		policyDefaultEA string
+		expectedEA      string
+	}{
+		{
+			policyConf: types.PolicyConfig{
+				Name: "policy-gatekeeper",
+				Manifests: []types.Manifest{
+					{
+						Path:              path.Join(tmpDir, "gatekeeper.yaml"),
```
mprahl, line 1815: Could you reuse `gatekeeperPath` here and also in other places?

internal/plugin_test.go
```diff
@@ -1760,6 +1762,162 @@ spec:
...
+						GatekeeperOptions: types.GatekeeperOptions{GatekeeperEnforcementAction: "deny"},
+					},
+				},
+			},
+			policyDefaultEA: "",
+			expectedEA: `
+            spec:
+                enforcementAction: deny
```
mprahl, line 1823: It'd be cleaner if you just provided `deny` here and the test code can do the work of checking for `spec` and `enforcementAction` being in the generated string.

internal/plugin_test.go
yiraeChristineKim: Good!

internal/plugin_test.go
yiraeChristineKim: oh Great

internal/utils.go
```diff
@@ -310,6 +316,23 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 	return policyTemplates, nil
 }
 
+// setGatekeeperEnforcementAction function override gatekeeper.constraint.enforcementAction
+func setGatekeeperEnforcementAction(manifest map[string]interface{}, enforcementAction string) error {
+	apiVersion, found, err := unstructured.NestedString(manifest, "apiVersion")
```
mprahl, line 321: Could you return right away if `enforcementAction == ""`?

internal/utils.go
```diff
@@ -310,6 +316,23 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
...
+	}
+
+	if found && strings.HasPrefix(apiVersion, "constraints.gatekeeper.sh") && enforcementAction != "" {
+		err := unstructured.SetNestedField(manifest, enforcementAction, "spec", "enforcementAction")
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
mprahl, line 333:
```suggestion
	if enforcementAction == "" {
		return nil
	}

	apiVersion, _, _ := unstructured.NestedString(manifest, "apiVersion")

	if strings.HasPrefix(apiVersion, "constraints.gatekeeper.sh") {
		err := unstructured.SetNestedField(manifest, enforcementAction, "spec", "enforcementAction")
		if err != nil {
			return err
		}
	}

	return nil
```

internal/utils.go
```diff
@@ -178,6 +178,12 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 		extraDeps := policyConf.Manifests[i].ExtraDependencies
 
 		for _, manifest := range manifestGroup {
+			err := setGatekeeperEnforcementAction(manifest,
+				policyConf.Manifests[i].GatekeeperEnforcementAction)
+			if err != nil {
+				return nil, err
```
mprahl, line 184: Could you provide more context in the return error such as the manifest index and that this error occurred due to setting `spec.enforcementAction`?

internal/utils.go
yiraeChristineKim: I think it would be more safe that we have error check here 
`apiVersion, _, err := unstructured.NestedString(manifest, "apiVersion")
if err != nil {
			return err
		}
`

mprahl: @yiraeChristineKim that's fine but I kind of wanted to avoid validating the actual manifest in the Policy Generator and let the policy controllers handle it.

mprahl (approved):

PR #169, opened 2024-07-16
Add setting name on ConfigurationPolicy

yiraeChristineKim:
When creating a policy with multiple manifests and `consolidateManfifest: false` the generated ConfigurationPolicies are names after the Policy with an index appended.

This causes issues when attempting to use extraDependencies such as below.  Adding an additional manifest will change the index causing the existing dependencies to be incorrect.

Allowing setting the name would also make it easier to identify in the UI when reviewing a policy.

Ref: https://issues.redhat.com/browse/ACM-12563

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -177,7 +177,7 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 		ignorePending := policyConf.Manifests[i].IgnorePending
 		extraDeps := policyConf.Manifests[i].ExtraDependencies
 
-		for _, manifest := range manifestGroup {
+		for j, manifest := range manifestGroup {
 			err := setGatekeeperEnforcementAction(manifest,
 				policyConf.Manifests[i].GatekeeperEnforcementAction)
 			if err != nil {
@@ -202,9 +202,10 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 				if found {
 					policyTemplate = buildPolicyTemplate(
 						policyConf,
-						len(policyTemplates)+1,
 						manifest["object-templates-raw"],
 						&policyConf.Manifests[i].ConfigurationPolicyOptions,
+						getConfigurationPolicyName(policyConf, i, len(policyTemplates),
+							j, !policyConf.ConsolidateManifests),
 					)
 				} else {
 					policyTemplate = map[string]interface{}{"objectDefinition": manifest}
@@ -256,9 +257,10 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 				// build policyTemplate for each objectTemplates
 				policyTemplate := buildPolicyTemplate(
 					policyConf,
-					len(policyTemplates)+1,
 					[]map[string]interface{}{objTemplate},
 					&policyConf.Manifests[i].ConfigurationPolicyOptions,
+					getConfigurationPolicyName(policyConf, i, len(policyTemplates),
+						j, !policyConf.ConsolidateManifests),
 				)
 
 				setTemplateOptions(policyTemplate, ignorePending, extraDeps)
@@ -279,9 +281,9 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 	if policyConf.ConsolidateManifests && len(objectTemplates) > 0 {
 		policyTemplate := buildPolicyTemplate(
 			policyConf,
-			1,
 			objectTemplates,
 			&policyConf.ConfigurationPolicyOptions,
+			getConfigurationPolicyName(policyConf, 0, 0, 0, false),
 		)
 		setTemplateOptions(policyTemplate, policyConf.IgnorePending, policyConf.ExtraDependencies)
 		policyTemplates = append(policyTemplates, policyTemplate)
@@ -315,6 +317,32 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 	}
 
 	return policyTemplates, nil
+}
+
+// This function returns configurationPolicyName with Index+1 based on the provided PolicyConfig.
+// When index is 0, it won't attach any number. When useManifestName is true,
+// it returns the policyConf.Manifests[manifestGroupIndex].Name otherwise,
+// It returns policyConf.Name + index
+func getConfigurationPolicyName(policyConf *types.PolicyConfig, manifestGroupIndex, policyNum,
+	manifestIndex int, useManifestName bool,
+) string {
+	// Do not append a number to configName when it is used for the first time.
+	configName := policyConf.Manifests[manifestGroupIndex].Name
+
+	if useManifestName && configName != "" {
+		if manifestIndex > 0 {
+			return fmt.Sprintf("%s%d", configName, manifestIndex+1)
+		}
+
+		return configName
+	}
+
+	configName = policyConf.Name
+	if policyNum > 0 {
+		return fmt.Sprintf("%s%d", configName, policyNum+1)
+	}
+
+	return configName
 }
 
 // setGatekeeperEnforcementAction function override gatekeeper.constraint.enforcementAction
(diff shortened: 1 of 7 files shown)


internal/utils.go
```diff
@@ -205,6 +205,7 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 						len(policyTemplates)+1,
 						manifest["object-templates-raw"],
 						&policyConf.Manifests[i].ConfigurationPolicyOptions,
+						"",
```
dhaiducek, line 208: I think we still need to detect whether a name is provided alongside the `object-templates-raw` manifest and add it here.

internal/utils.go
```diff
@@ -252,13 +253,20 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
...
+				}
+
 				// casting each objTemplate with manifest to objectTemplates type
 				// build policyTemplate for each objectTemplates
 				policyTemplate := buildPolicyTemplate(
 					policyConf,
 					len(policyTemplates)+1,
 					[]map[string]interface{}{objTemplate},
 					&policyConf.Manifests[i].ConfigurationPolicyOptions,
+					configName,
```
dhaiducek, line 269: (Optional) I might prefer we pass the parent object to the function and set `policyConf.Manifests[i].Name` in `ApplyDefaults()`: 
```suggestion
					&policyConf.Manifests[i],
```
Then in the conditional above, we detect whether `policyConf.name == Manifests[i].Name` to determine whether it's been overridden, and then set `Manifests[i].Name` directly with the appropriate index if necessary and use that value directly for the name in `buildPolicyTemplate()`. This would also properly handle the case where the same name is given for the manifest and the policy.

dhaiducek (changes requested):
Thanks for the update! I have some comments for consideration.

internal/utils.go
yiraeChristineKim: `object-templates-raw` is isPolicyTypeManifest = true so it should have `name` otherwise it throws an error  at `isPolicyTypeManifest, isOcmPolicy, err := isPolicyTypeManifest(` line 188. Do we need to add the name there?

internal/utils.go
yiraeChristineKim: I need a` j` (When it has multiple manifest in the Path file) to attach numbers to configurationpolicy name

dhaiducek: Right, I'm proposing the conditional above these lines would use `j` and set the generator struct directly before passing it to `buildPolicyTemplate()`, something like this to detect whether the name was overridden:
```
if j != 0 && policyConf.name != policyConf.Manifests[i].Name {
	policyConf.Manifests[i].Name = fmt.Sprintf("%s%d", policyConf.Manifests[i].Name, j+1)
}
```

internal/utils.go
yiraeChristineKim: configurationPolicy name?

internal/utils.go
yiraeChristineKim: I refactored. I added this func
```
func getConfigurationPolicyName(policyConf *types.PolicyConfig, manifestGroupIndex, policyNum,
	manifestIndex int, useManifestName bool,
) string {
	// Do not append a number to configName when it is used for the first time.
	var configName string

	if useManifestName && policyConf.Manifests[manifestGroupIndex].Name != "" {
		// If the user does not added Manifests[manifestGroupIndex].Name
		// then return policyConf.Name as a configurationPolicyName
		configName = policyConf.Manifests[manifestGroupIndex].Name
		if manifestIndex > 0 && configName != "" {
			return fmt.Sprintf("%s%d", configName, manifestIndex+1)
		}
	}

	if configName == "" {
		configName = policyConf.Name
		if policyNum > 1 {
			return fmt.Sprintf("%s%d", configName, policyNum)
		}
	}

	return configName
}
```

internal/utils.go
dhaiducek: The ConfigurationPolicy name is customizable in the generator manifest--it won't be set in the manifest in this scenario.

internal/utils.go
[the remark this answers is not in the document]
dhaiducek: I was thinking the loop would have its own counter that only increments if the name wasn't overridden. So, copying/pasting my other sample, using something like a `templateCounter`:
```
if j != 0 && policyConf.name != policyConf.Manifests[i].Name {
	policyConf.Manifests[i].Name = fmt.Sprintf("%s%d", policyConf.Manifests[i].Name, j+1)
} else {
  templateCounter++
}
```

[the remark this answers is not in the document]
yiraeChristineKim: Why templateCounter increase when j == 0?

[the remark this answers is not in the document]
dhaiducek: Oh, good point. So it'd be something like this then?
```golang
if j != 0 && policyConf.name != policyConf.Manifests[i].Name {
	policyConf.Manifests[i].Name = fmt.Sprintf("%s%d", policyConf.Manifests[i].Name, j+1)
} else if policyConf.name == policyConf.Manifests[i].Name {
  templateCounter++
}

```

internal/utils.go
yiraeChristineKim: @Dale I added a unit test for `object-templates-raw` When you are free, When you have a moment, could you review it to ensure we're on the same page?

dhaiducek: Looks good, thanks! I dug into this. It looks like tacking on the index of the policy-template is already the current behavior (for example, providing a ConfigurationPolicy directly would bring its name similar to how this feature does) and adding a counter like I'm proposing would be too different.

internal/utils.go
```diff
@@ -254,11 +255,13 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
 			} else {
 				// casting each objTemplate with manifest to objectTemplates type
 				// build policyTemplate for each objectTemplates
+
```
dhaiducek, line 258: nit: extra newline
```suggestion
```

internal/utils.go
```diff
@@ -317,6 +320,35 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
...
+	if configName == "" {
+		configName = policyConf.Name
+		if policyNum > 1 {
+			return fmt.Sprintf("%s%d", configName, policyNum)
+		}
+	}
+
+	return configName
+}
```
dhaiducek, line 350: (optional) `configName` could be assigned to `policyConf.Manifests[manifestGroupIndex].Name` up front. The flow might be easier to understand that way.
```suggestion
	// Do not append a number to configName when it is used for the first time.
	configName := policyConf.Manifests[manifestGroupIndex].Name

	if useManifestName && configName != "" {
		if manifestIndex > 0 {
			return fmt.Sprintf("%s%d", configName, manifestIndex+1)
		}

		return configName
	}

	// Fall back to policy name
	configName = policyConf.Name
	if policyNum > 1 {
		return fmt.Sprintf("%s%d", configName, policyNum)
	}

	return configName
}
```

internal/utils.go
```diff
@@ -317,6 +320,35 @@ func getPolicyTemplates(policyConf *types.PolicyConfig) ([]map[string]interface{
...
+		configName = policyConf.Manifests[manifestGroupIndex].Name
+		if manifestIndex > 0 && configName != "" {
+			return fmt.Sprintf("%s%d", configName, manifestIndex+1)
+		}
+	}
+
+	if configName == "" {
+		configName = policyConf.Name
+		if policyNum > 1 {
+			return fmt.Sprintf("%s%d", configName, policyNum)
```
dhaiducek, line 345: (optional) With the helper function, we could move the increment logic here (and it'd match the manifest code above for consistency).
```suggestion
		if policyNum > 0 {
			return fmt.Sprintf("%s%d", configName, policyNum+1)
```

dhaiducek (approved):
/hold

I have a bunch of optional comments, so I've approved with a hold if you want to address them but overall looks good to me!

internal/utils.go
yiraeChristineKim: good!

dhaiducek (approved):

PR #176, opened 2024-08-21
Add support for custom compliance messages in ConfigurationPolicy

zyjjay:
The new `.spec.customMesssage.Compliant` and `spec.customMessage.noncompliant` fields of ConfigurationPolicy are now supported. Note that when consolidateManifests is set to true, the custom message configuration should be set at the policy level and not at the manifest level.

ref: https://issues.redhat.com/browse/ACM-13134

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -342,6 +342,24 @@ func isEvaluationIntervalSet(config map[string]interface{}, policyIndex int, com
 	return set
 }
 
+// isCustomMessageSet will return whether the customMessage field of the specified policy is set in the
+// Policy Generator configuration YAML
+func isCustomMessageSet(config map[string]interface{}, policyIndex int, complianceType string) bool {
+	policy := getPolicy(config, policyIndex)
+	if policy == nil {
+		return false
+	}
+
+	customMessage, ok := policy["customMessage"].(map[string]interface{})
+	if !ok {
+		return false
+	}
+
+	_, set := customMessage[complianceType].(string)
+
+	return set
+}
+
 // isEvaluationIntervalSetManifest will return whether the evaluation interval of the specified manifest
 // of the specified policy is set in the Policy Generator configuration YAML.
 func isEvaluationIntervalSetManifest(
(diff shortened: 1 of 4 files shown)


mprahl:
@zyjjay it seems the inheritance isn't taken into account (e.g. if set on policyDefaults but on the policy, policyDefaults is used). Take a look at the `applyDefaults` function and account for this new field.

zyjjay:
Actually I was thinking if the `customMessages` field is unset in the generated policy, the default message templating will be handled by the config policy controller.

internal/plugin.go
```diff
@@ -1079,6 +1079,10 @@ func (p *Plugin) assertValidConfig() error {
 					return fmt.Errorf(errorMsgFmt, "severity")
 				}
 
+				if manifest.CustomMessage != nil {
```
dhaiducek, line 1082: This is fine if the messages are the same.
```suggestion
				if !reflect.DeepEqual(manifest.CustomMessage, policy.CustomMessage) {
```

internal/types/types.go
```diff
@@ -45,6 +45,7 @@ type ConfigurationPolicyOptions struct {
 	PruneObjectBehavior    string             `json:"pruneObjectBehavior,omitempty" yaml:"pruneObjectBehavior,omitempty"`
 	RecordDiff             string             `json:"recordDiff,omitempty" yaml:"recordDiff,omitempty"`
 	RecreateOption         string             `json:"recreateOption,omitempty" yaml:"recreateOption,omitempty"`
+	CustomMessage          *CustomMessage     `json:"customMessage,omitempty" yaml:"customMessage,omitempty"`
```
dhaiducek, line 48: Similar to `EvaluationInterval`, I'm not sure there's a benefit to making this a pointer--unless there was a reason to do so?
```suggestion
	CustomMessage          CustomMessage     `json:"customMessage,omitempty" yaml:"customMessage,omitempty"`
```

internal/utils.go
```diff
@@ -487,11 +487,22 @@ func buildPolicyTemplate(
 	configPolicyOptionsOverrides *types.ConfigurationPolicyOptions,
 	configPolicyName string,
 ) map[string]interface{} {
+	var customMessage *types.CustomMessage
+	if policyConf.ConsolidateManifests {
+		customMessage = policyConf.CustomMessage
+	} else {
+		customMessage = configPolicyOptionsOverrides.CustomMessage
+	}
+
```
dhaiducek, line 496: This logic should live in the `applyDefaults()` function.

internal/utils.go
```diff
@@ -487,11 +487,22 @@ func buildPolicyTemplate(
...
 	policySpec := map[string]interface{}{
 		"remediationAction": policyConf.RemediationAction,
 		"severity":          policyConf.Severity,
 	}
 
+	if customMessage != nil {
+		policySpec["customMessage"] = customMessage
+	}
+
```
dhaiducek, line 505: This is oversimplified. It should look more like the `evaluationInterval` conditional further down.

mprahl:
Superseded by #178

PR #177, opened 2024-08-27
Add support for spec.hubTemplateOptions when generating a policy

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-13608

### DIFF
--- a/internal/types/types.go
+++ b/internal/types/types.go
@@ -7,6 +7,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type HubTemplateOptions struct {
+	ServiceAccountName string `json:"serviceAccountName,omitempty" yaml:"serviceAccountName,omitempty"`
+}
+
 type PolicyOptions struct {
 	Categories                     []string           `json:"categories,omitempty" yaml:"categories,omitempty"`
 	Controls                       []string           `json:"controls,omitempty" yaml:"controls,omitempty"`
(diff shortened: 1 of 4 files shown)


internal/types/types.go
```diff
@@ -7,27 +7,32 @@ import (
...
+	InformGatekeeperPolicies       bool                `json:"informGatekeeperPolicies,omitempty" yaml:"informGatekeeperPolicies,omitempty"`
+	InformKyvernoPolicies          bool                `json:"informKyvernoPolicies,omitempty" yaml:"informKyvernoPolicies,omitempty"`
+	GeneratePolicyPlacement        bool                `json:"generatePolicyPlacement,omitempty" yaml:"generatePolicyPlacement,omitempty"`
+	GeneratePlacementWhenInSet     bool                `json:"generatePlacementWhenInSet,omitempty" yaml:"generatePlacementWhenInSet,omitempty"`
+	PolicySets                     []string            `json:"policySets,omitempty" yaml:"policySets,omitempty"`
+	PolicyAnnotations              map[string]string   `json:"policyAnnotations,omitempty" yaml:"policyAnnotations,omitempty"`
+	PolicyLabels                   map[string]string   `json:"policyLabels,omitempty" yaml:"policyLabels,omitempty"`
+	ConfigurationPolicyAnnotations map[string]string   `json:"configurationPolicyAnnotations,omitempty" yaml:"configurationPolicyAnnotations,omitempty"`
+	HubTemplateOptions             *HubTemplateOptions `json:"hubTemplateOptions,omitempty" yaml:"hubTemplateOptions,omitempty"`
```
dhaiducek, line 35: Looks like this pointer is copy/paste, but does it need to be a pointer when it has `omitempty`?

dhaiducek (approved):
/hold

LGTM! I have one question about the pointer. It seems to me the logic might be simpler without it (though it's probably fine if we're trying to be consistent with the propagator).

internal/types/types.go
mprahl: It doesn't have to be and seems to work well without it being a pointer. Updated!

dhaiducek (approved):

PR #178, opened 2024-08-28
Add setting custom messages on configuration policies

mprahl:
This supersedes #176 and the tests were taken from there also.

Relates:
https://issues.redhat.com/browse/ACM-13134

### DIFF
--- a/internal/plugin.go
+++ b/internal/plugin.go
@@ -324,15 +324,13 @@ func getPolicySet(config map[string]interface{}, policySetIndex int) map[string]
 	return getArrayObject(config, "policySets", policySetIndex)
 }
 
-// getEvaluationInterval will return the evaluation interval of specified policy in the Policy Generator configuration
-// YAML.
-func isEvaluationIntervalSet(config map[string]interface{}, policyIndex int, complianceType string) bool {
+func isComplianceConfigSet(config map[string]interface{}, policyIndex int, field, complianceType string) bool {
 	policy := getPolicy(config, policyIndex)
 	if policy == nil {
 		return false
 	}
 
-	evaluationInterval, ok := policy["evaluationInterval"].(map[string]interface{})
+	evaluationInterval, ok := policy[field].(map[string]interface{})
 	if !ok {
 		return false
 	}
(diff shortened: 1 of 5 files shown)


internal/plugin.go
```diff
@@ -700,6 +700,14 @@ func (p *Plugin) applyDefaults(unmarshaledConfig map[string]interface{}) {
 			policy.HubTemplateOptions.ServiceAccountName = p.PolicyDefaults.HubTemplateOptions.ServiceAccountName
 		}
 
+		if policy.CustomMessage.Compliant == "" {
+			policy.CustomMessage.Compliant = p.PolicyDefaults.CustomMessage.Compliant
```
dhaiducek, line 704: Should this use something similar to (or just reuse) the `isEvaluationIntervalSet()`?
```suggestion
			policy.CustomMessage.Compliant = p.PolicyDefaults.CustomMessage.Compliant
```

internal/utils.go
```diff
@@ -492,6 +492,11 @@ func buildPolicyTemplate(
 		"severity":          policyConf.Severity,
 	}
 
+	if configPolicyOptionsOverrides.CustomMessage.Compliant != "" ||
+		configPolicyOptionsOverrides.CustomMessage.NonCompliant != "" {
+		policySpec["customMessage"] = configPolicyOptionsOverrides.CustomMessage
+	}
```
dhaiducek, line 498: I think it'd make sense to handle this similar to `evaluationInterval` where `Compliant` and `NonCompliant` are handled separately. I could see a case where a user wants to have a default brief message for everything and then override just `NonCompliant` for one policy (or vice versa).

(and probably move this block adjacent to it since they're both similar settings--though I'm not entirely sure how the `policyTemplate` wound up in the middle of this function...)

mprahl: I think due to the omitempty in the YAML tag on the struct, it ends up doing the same thing but let's make it consistent with evaluation interval.

internal/plugin.go
mprahl: Good point. That is more flexible.

dhaiducek (approved):
LGTM! Thanks!

PR #184, opened 2024-09-20
Document the watch evaluation interval

mprahl:


docs/policygenerator-reference.yaml
```diff
@@ -60,14 +60,15 @@ policyDefaults:
...
-  # clusters have low CPU resources, the evaluation interval can be increased to to reduce CPU usage on the Kubernetes
-  # API.
+  # Optional. Configures the minimum elapsed time before a configuration policy is reevaluated. The default value is
+  # `watch` to leverage Kubernetes API watches instead of polling the Kubernetes API server. If the policy spec is
+  # changed or if the list of namespaces selected by the policy changes, the policy might be evaluated regardless of the
+  # settings here.
   evaluationInterval:
     # These are in the format of durations (e.g. "1h25m3s"). These can also be set to "never" to avoid evaluating the
-    # policy after it has become a particular compliance state.
+    # policy after it has become a particular compliance state. Default is `watch`.
```
dhaiducek, line 69: (Optional) Clarify since there's an example of a duration here.
```suggestion
    # policy after it has become a particular compliance state. The default value for both fields is `watch`.
```

dhaiducek (approved):

Issue #187, opened 2024-09-26
bug - gatekeeper expander - wrong event_type set for generated inform-gatekeeper-admission-* ConfigurationPolicy


itewk:
# Context
According to the gatekeeper docs (https://open-policy-agent.github.io/gatekeeper/website/docs/audit/#audit-logs) and experimentation, the Events logged by Gatekeeper have an `event_type: violation_audited`

# Problem
The current gatekeeper expander generates a ConfigurationPolicy looking for events of `event_type: violation` which it will never find.
https://github.com/open-cluster-management-io/policy-generator-plugin/blob/8baf967b30b5e8f765fa755fd08f0a1a8c1e7c64/internal/expanders/gatekeeper.go#L113

# solution
change the generation to `event_type: violation_audited`

PR #188, opened 2024-09-26
Issue #187 - bug - gatekeeper expander - wrong event_type set for generated inform-gatekeeper-admission-* ConfigurationPolicy

itewk:
Resolves issue #187

### DIFF
--- a/internal/expanders/gatekeeper.go
+++ b/internal/expanders/gatekeeper.go
@@ -110,7 +110,7 @@ func (g GatekeeperPolicyExpander) Expand(
 								"constraint_action": "deny",
 								"constraint_kind":   constraintKind,
 								"constraint_name":   constraintName,
-								"event_type":        "violation",
+								"event_type":        "violation_audited",
 							},
 						},
 					},
--- a/internal/expanders/gatekeeper_test.go
+++ b/internal/expanders/gatekeeper_test.go
@@ -145,7 +145,7 @@ func TestGatekeeperExpand(t *testing.T) {
 									"constraint_action": "deny",
 									"constraint_kind":   "MyConstraint",
 									"constraint_name":   "my-awesome-constraint",
-									"event_type":        "violation",
+									"event_type":        "violation_audited",
 								},
 							},
 						},


mprahl (approved):

dhaiducek (changes requested):
@itewk It's not actually incorrect. The event_type for admission events is indeed `violation`:

https://github.com/open-policy-agent/gatekeeper/blob/0a4851a975d25003218a23b8a70217f759f2454f/pkg/webhook/policy.go#L297

However, we'd welcome a contribution to add the `violation_audit` to the expander.

mprahl (changes requested):
Thanks for catching that @dhaiducek!

internal/expanders/gatekeeper.go
```diff
@@ -110,7 +110,7 @@ func (g GatekeeperPolicyExpander) Expand(
 								"constraint_action": "deny",
 								"constraint_kind":   constraintKind,
 								"constraint_name":   constraintName,
-								"event_type":        "violation",
+								"event_type":        "violation_audited",
 							},
 						},
 					},
```
dhaiducek, line 116: The simplest thing might be to just add it as another `object-template`. What would you think of that, @itewk?
```suggestion
					},
					{
						"complianceType": "mustnothave",
						"objectDefinition": map[string]interface{}{
							"apiVersion": "v1",
							"kind":       "Event",
							"annotations": map[string]interface{}{
								"constraint_action": "deny",
								"constraint_kind":   constraintKind,
								"constraint_name":   constraintName,
								"event_type":        "violation",
								"event_type":        "violation_audited",
							},
						},
					},
```

itewk:
aaah. okay.

internal/expanders/gatekeeper.go
itewk: yeah. but im also fighting with the hard coded namespaceSelector and the note left behind about "futer improvements needed". its annoying having namepsaces /excludedNamespaces set in my constraint (both at the gatekeeper and OPM level) but then those don't propigate to the generated ConfigurationPolicy(s) which ar harded cluded eclude/includes.

i am going to keep playing with it to see if i can get the behavor i would expect.

PR #189, opened 2024-09-27
Gatekeeper Expander - add audit event inform ConfigurationPolicy and add ability for the inform ConfiguraitonPolicies to read the Constraints included and excluded namespaces rather then hard coded defaults

itewk:
Sorry for two things in one. Can try to separate them out if really needed.

* Adds inform ConfiguraitonPolicy for audit events (and renames the existing inform ConfigurationPolicies for clarity)

* Adds feature for the included and excluded namespaces for the generated inform ConfiguraitonPolicies to be read from the Constraints include and exclude namespaces so the inform is at the same scope as the Constraint


itewk:
@gparvin and @mprahl  thoughts?

itewk:
now that i understand this flag is deprecated and why ill just close this

PR #196, opened 2025-02-26
Add `objectSelector`

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-18011


gparvin (approved):
This makes sense to me.

dhaiducek:
/unhold

Did some cursory manual tests, and it seems to be working. One thing I will note (that also applies to the `namespaceSelector`) is that neither selector can be overridden to erase it--only override its fields. We should probably address that in the future.

PR #251, opened 2026-03-16
Expand new kyverno policy types

jan-law:
Support policy generator expansion for the validating, mutating, generating, and image validation policy types in policies.kyverno.io/v1

ref: https://issues.redhat.com/browse/ACM-30858


JustinKuli (approved):
LGTM!

